PR #45, opened 2015-07-25
Use [BsonField] to serialize non-public property

techrising:
Thank you for your awesome project ! It's very well design and very fast.
I just add the feature to serialize an internal property via Attribute [BsonProperty].
Moreover it could be nice if you sign the dll. I will be able to use it in nuget ;)
Thanks again for your work
Kind Regards


mbdavid:
Hi @techrising, thanks for your contribuition. I have some questions:

- Why not use `BsonFieldAttribute` that exists? In LiteDB i try to use definition names like "Classes has Properties, Document (Bson) has Fields" - So when I refer to a Class I use "Property", when is Document I use "Field"

- By default, internal/protected/private properties could not be serialized - only if has `[BsonField]`

techrising:
Hi @mbdavid 

> Why not use BsonFieldAttribute that exists? 

- You are perfectly right for the name.I just didn't want to change the implementation of the bson field because it must have a "Name" and that's not something required.
If you don't have a problem with [BsonField] I can change it and use it.

> By default, internal/protected/private properties could not be serialized - only if has [BsonField]

- Yes only if they have the attribute. I already add few tests (on Mapper_test).
If you want to add this feature on the next release I will create a dedicated test method.

let me know your thoughts and I will do it.

mbdavid:
Nice works! I'm merging with master branch and will release this feature with 1.1 version that I'm working on it. Thanks

techrising:
Thanks! I'm looking forward to try it.
And please sign the dll on nuget ;)

PR #47, opened 2015-07-26
Added DropCollection unit test.

rdstevens:
Here's a trivial DropCollection unit test, if you want it!


mbdavid:
I will made some tests now.

PR #63, opened 2015-09-21
Features/foreach

tbdbj:
add IEnumerable<T> feature to LiteCollection.
Using is more native:
foreach(var item in collection) ...
instead foreach(var item in collection.FindAll()) ...


mbdavid:
Hi @tbdbj, can be more native to use but has some drawbacks:

- Can be used only when you need all objects
- If you try filter using Linq, you will not use index search (will be a full scan search), like:
```
// Full scan - Very Slow
foreach(var user in users.Where(x => x.UserName == "John")) {

// Index search - Instant
foreach(var user in users.Find(x => x.UserName == "John")) {
```

PR #65, opened 2015-09-21
_writer => Lazy<BinaryWriter>

tbdbj:
change _writer from BinaryWriter to Lazy<BinaryWriter>


mbdavid:
Has a problem in this implementation, every instance method `WritePage` call will re-open filestream, becase:

```
        /// <summary>
        /// Write a page from memory to disk 
        /// </summary>
        public void WritePage(BasePage page)
        {
            DiskService.WritePage(GetWriter(), page);
        }
```

PR #68, opened 2015-10-07
fix bugs on drop and delete commands

alanlima:
Hi guys,

I found some issues when I needed delete some data from the database.
When I use the command delete, the result is the error "The given key was not present in the dictionary".
I also tried use the command drop, but resulted in the same error.

See the printscreen:

2015-10-07 12_35_43-litedb running - microsoft visual studio

I forked the repository and fix the issues by adding some validation check in the IndexServices.

I corrected them and now I can execute the delete and drop commands well, and the find command too.

See the printscreen:

2015-10-07 12_38_14-litedb running - microsoft visual studio

### DIFF
--- a/LiteDB/Storage/Services/IndexService.cs
+++ b/LiteDB/Storage/Services/IndexService.cs
@@ -235,7 +235,7 @@ namespace LiteDB
         {
             if (address.IsEmpty) return null;
             var page = _pager.GetPage<IndexPage>(address.PageID);
-            return page.Nodes[address.Index];
+            return page.Nodes.GetOrDefault(address.Index, null);// page.Nodes[address.Index];
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


mbdavid:
Hi @alanlima, thanks for your contribuition. The question is: why this error happing (could be a data corruped database)? Do you have steps to create your `data.db`?

Do a simple test: in shell, after delete/drop your collection, run `dump` command to know how pages are.

I'm in vacations now and have no computer to test now (back next week).

alanlima:
Hi @mbdavid, I understand the questions but I no have the steps to create the data.db.

I use the database for a simple windows service to log and storage some data. The service can run in multiple tasks and write in the database at the same time. That is the case of the executionlog document. Maybe this is the problem...

I'm not sure how I could analyse the pages, here a printscreen from result of dump command:

2015-10-08 17_21_20-start

If you need the data.db file, I can send you without problems.

Please, if you need some more help, I will be pleased in help you.

mbdavid:
Hi @alanlima, I´m back. Let me ask something: did you re-use same `LiteDatabase` instance? Remember that LiteDB is not thread safe, you must always use LiteDatabase inside `using` statment.
If is not your case, can you send me your database to analyse? Thanks

PR #88, opened 2015-12-09
Added ability to use GetCollection<T>() without collection name. this…

SonOfSardaar:
… assigns collection name automatically. for.e.g for object 'Customer' it will assign collection name to 'CustomerCollection'

Also i have used 'FluentAssertions' library to assert the results easily.

### DIFF
--- a/LiteDB.Tests/ImplicitCollectionName/BasicTestsOnImplicitCollectionNames.cs
+++ b/LiteDB.Tests/ImplicitCollectionName/BasicTestsOnImplicitCollectionNames.cs
@@ -1,0 +1,62 @@
+﻿using System.IO;
+using FluentAssertions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace LiteDB.Tests.ImplicitCollectionName
+{
+    [TestClass]
+    public class BasicTestsOnImplicitCollectionNames
+    {
+        public class BasicObject
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+        }
+
+        [TestMethod]
+        public void WillAssignCollectionNameIfNotSpecified()
+        {
+            using (var stream = new MemoryStream())
+            {
+                using (var db = new LiteDatabase(stream))
+                {
+                    var objects = db.GetCollection<BasicObject>();
+                    objects.Insert(new BasicObject());
+                }
+
+                using (var db = new LiteDatabase(stream))
+                {
+                    var actualCollectionNames = db.GetCollectionNames();
+                    var expectedCollectionNames = new[] {"BasicObjectCollection"};
+
+                    actualCollectionNames.ShouldAllBeEquivalentTo(expectedCollectionNames);
+                }
+            }
+            
+        }
+
+        [TestMethod]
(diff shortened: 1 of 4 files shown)


Skysper:
This fetcher is so nice. with this, I need not to set the collection name by the T object's type name.But I do think it should use the type's full name without dots.

mbdavid:
Hi guys, `GetCollection` without an explicit name was used in very first version of LiteDB and removed after. 

Has some problems to LiteDB assume collections name, like:

- I can use `GetCollection<BsonDocument>` that returns any collection without strong type.
- Different classes can be same name (diferent assembly/namespace) - ok, can be used Assembly full name, but collection name will be very long

LiteDB was design to work without any custom strong entity. So, with you write an app with a strong entity I always can open in shell app, for example. Another application don't know your entities classes (like Mongo collections or any table in RMDBS).

I think this method can be used as an `Extension` method or a method from a class that extend LiteDatabase. Or maybe an external CONTRIB project?

Of course, it's an open source project and I'm open from another ideias. What did you thing about?

SonOfSardaar:
Hi Guys
In my opinion we have both options with this.
Anyone wants to use collection name explicitly can use it too. Secondly in real world I doubt you will try to use same class name in data layer. So we can use pluralized class name for collection. But to avoid another dependency I post fixed collection to class name.
Regards
Gurpreet Singh
On 13 Dec 2015 11:42 pm, Mauricio David <notifications@github.com> wrote:Hi guys, GetCollection without an explicit name was used in very first version of LiteDB and removed after. 

Has some problems to LiteDB assume collections name, like:

I can use GetCollection<BsonDocument> that returns any collection without strong type.Different classes can be same name (diferent assembly/namespace) - ok, can be used Assembly full name, but collection name will be very long

LiteDB was design to work without any custom strong entity. So, with you write an app with a strong entity I always can open in shell app, for example. Another application don't know your entities classes (like Mongo collections or any table in RMDBS).

I think this method can be used as an Extension method or a method from a class that extend LiteDatabase. Or maybe an external CONTRIB project?

Of course, it's an open source project and I'm open from another ideias. What did you thing about?

techrising:
The idea is good, but I prefer to have it in an external project.
Moreover LiteDB is a single dll project and It's better to keep it like that.

SonOfSardaar:
I am OK with separate project if that's all think. Only problem is that people will have to create new repos and projects for one or two useful extensions
Regards
Gurpreet Singh
On 14 Dec 2015 8:39 am, Kevin Gallerin <notifications@github.com> wrote:The idea is good, but I prefer to have it in an external project.
Moreover LiteDB is a single dll project and It's better to keep it like that.

SonOfSardaar:
Moreover if the idea is good then its better to keep it in core to keep it single assembly
Regards
Gurpreet Singh
On 14 Dec 2015 8:39 am, Kevin Gallerin <notifications@github.com> wrote:The idea is good, but I prefer to have it in an external project.
Moreover LiteDB is a single dll project and It's better to keep it like that.

Skysper:
I agree that just leave the LiteDB what it is now.I have not got a better idea about this.Just let people implement it themselves.

Issue #92, opened 2015-12-30
Add support for deserialization of IEnumerable


michal-novak-pos-digital:
Please add support for deserialization of IEnumerable in similar way as XmlSerializer does.
See remarks about IEnumerable https://msdn.microsoft.com/en-us/library/system.xml.serialization.xmlserializer(v=vs.110).aspx

It would be great to keep compatible behavior.
Anyway this method crashes when target collection is derived from List<T>

You can use attached implemetation.
[Deseriaize.txt](https://github.com/mbdavid/LiteDB/files/75030/Deseriaize.txt)

mbdavid:
Hi @michal-novak-pos-digital, thanks for your contribution. I use your implementation for IEnumerable changing the way how get itemType (your way doest work on my unit tests). Can you test this last commit?

michal-novak-pos-digital:
Thank you very much for including this enhacement int your project.

But there still is one issue.
When type derived from IEnumerable<>, ICollection<>, or IList<>, etc.. is used deserialization fails.
Prease see my pull request.

My solution failed when IEnumerable<> was used directly, but yours failed as described above.
So this would make this problem solved for both cases.

PR #95, opened 2016-01-04
Added support for deserialization of types derived from IEnumerable<>

michal-novak-pos-digital:
Fixed issue when type derived from IEnumerable<>, ICollection<>, or IList<>, etc.. is used deserialization fails.

For details please see issue #92


mbdavid:
Hi @michal-novak-pos-digital, thanks for your PR. I will accept but will add your logic to get type in Reflection.cs class to be more easy to migrate to Portable version.

PR #103, opened 2016-01-15
Replace unnecessary LINQ.Count() usage

remember664:
In the following cases we can replace LINQ.Count() to improve performance.


mbdavid:
Nice. Thanks!

PR #106, opened 2016-01-24
fix overflow bug in TransactionService

remember664:
I faced with bug that LiteDB could not handle file with length in bytes more than `UInt.MaxValue`.

Then i found overflow issues in the `TransactionService` and in a few additional places.
P.S. Similar code without bug (with explicit `long` usage) has been updated as well for keeping code consistent.


mbdavid:
Hi @remember664, nice one. I never test with datafiles large than `UInt.MaxValue`.

PR #109, opened 2016-01-27
Unity3d

buronix:
Hi,

Im a Game Developer and I think your db is perfect for Unity (http://unity3d.com/), but the problem with Unity is that it only implement .Net 2.0 (and some things up to 3.5)  because it uses a custom mono version, but I manage to perform some small changes to make it works under Unity without any functionality or performance reduction in you base version.

It would me awesome to implement these changes or others similar (Maybe a isNullOrWhiteSpace String extension).

I think with a little publicity on the Forums this DB can be used a lot by Unity Developers.

PS: Awesome Job, Im using LiteDB and I think its awesome!!!.

Ruben.


mbdavid:
Hi @buronix, thanks for your PR. I saw your project but isnt still in v4? In LiteDB.csproj has <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>

I think that LiteDB use more api from v4.0... But your idea is fine, I upgrade to v4 to use Async but never apply that.

buronix:
Hi, Sorry, Unity is a plataform to develop Games, a lot of Famours Games are developed under Unity (http://unity3d.com/showcase/gallery), and tons of indie Games  also, http://www.indiedb.com/engines/unity/games, this is the reason for Ithink it will be awesome support Unity, But Im not from Unity, Im a Indie Game Developer, actually developing a game underUnity.

About the csproj, Im not using it, so I change nothing with that, I only imported the LiteDB folder and performed the changes to make it work under Unity, so Im not using a dll, (Maybe in the final release).

mbdavid:
Hi @buronix, I change myself last version of LiteDB to works in 3.5 framework. But Shell and Test must run in v4 (because v.1.0.4 is compiled in v4 and shell use v.0.9 and v.1.0.4 to be compatible with old versions).

buronix:
Thanks a Lot!!

Issue #112, opened 2016-01-28
Support Enumerable.All and Enumerable.Any


franzalex:
I propose that you include support for parsing the `System.Linq.Enumerable.All()` and `System.Linq.Enumerable.Any()` methods to the corresponding `QueryAnd` and `QueryOr` expressions. Doing this will allow convenient searching for records in which one field is compared with multiple values from a collection.  

An example of the code that can be written when this is implemented is demonstrated below:

```c#
using (var db = new LiteDatabase(new MemoryStream()))
{
    var col = db.GetCollection<User>("Users");

    col.EnsureIndex(x => x.Name, true);

    col.Insert(new[] { new User() { Id = 1, Name = "John Smith" },
                       new User() { Id = 2, Name = "Jane Smith" },
                       new User() { Id = 3, Name = "John Doe" },
                       new User() { Id = 4, Name = "Jane Doe" } });
    
    var names1 = new[] { "John", "Joe", "Jane", "Janet" };
    col.Find(user => names1.Any(name => user.Name.StartsWith(name))); // all four users
    
    var names2 = new[] { "John", "Doe" }
    col.Find(user => names2.All(name => user.Name.Contains(name))); // returns: John Doe
    col.Find(user => names2.Any(name => user.Name.Contains(name))); /* returns: John Doe 
                                                                     *          John Smith 
                                                                     *          Jane Doe
                                                                     */
 }
```

PR #113, opened 2016-01-28
Implemented querying collections with Enumerable.All and Enumerable.Any

franzalex:
Implemented feature request in issue #112.

### DIFF
--- a/LiteDB.Tests/Index/LinqTest.cs
+++ b/LiteDB.Tests/Index/LinqTest.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.IO;
+using System.Linq;
 
 namespace LiteDB.Tests
 {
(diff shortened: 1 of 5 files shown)


mbdavid:
Hi @franzalex, nice works. Thanks... But, I why not use `QueryIn` for `Enumerable.Any` and implement another `QueryAll` for `Enumerable.All`?

franzalex:
The limitation with the current implementation is that `QueryIn` uses only the `Object.Equals()` method exclusively while `QueryAll` returns all the records in a collection with no exceptions.

With this implementation, the the limitation I described above is removed, allowing you to match a field with any method you desire (i.e. `Contains`, `StartsWith`, `Equals`, `>`, `<=`, etc). The following code shows what one had to do before and what can be done now.

```c#
var col = db.GetCollection<Person>("people");
col.EnsureIndex(x => x.Name, true);

// Test 1:  find people whose names start with John, Jane, Joe or Judy
var anyNames = new[] { "John", "Jane", "Joe", "Judy" };

// Before my submission (i.e. no Enumerable.Any)
var query1 = anyNames.Select(name => Query.StartsWith("Name", name))
                     .Aggregate((lhs, rhs) => Query.Or(lhs, rhs));
col.Find(query1);

// With my submission
col.Find(person => anyNames.Any(name => person.Name.StartsWith(name));


// Test 2: Find persons whose names contain all of the following John, W. and Smith
var allNames = new[] { "John", "W.", "Smith" };

// Before my submission (i.e. no Enumerable.All)
var query2 = allNames.Select(name => Query.Contains("Name", name))
                     .Aggregate((lhs, rhs) => Query.And(lhs, rhs));
col.Find(query2);

// With my submission
col.Find(person => allNames.All(name => person.Name.Contains(name));
```

Yes, it is true, you can do the same things using both methods. I believe though that my additions provide more convenience through the use of lambda expressions.

mbdavid:
You are right. I never used `Any` or `All` enumetarion method, that why I didn´t undersand.. my `Linq` skill are terrible; Again, thanks for code and explanation. I will merge this weekend.

PR #118, opened 2016-02-03
Fixed mscorlib exception which arises from journalling not being enabled

jaked122:
I was a bit frustrated by the constant mscorlib exceptions arising from FileDiskServer::TryRecovery, so I added an if statement to check if the journal function is enabled.

I don't know if you want this fixed, but I find the lack of exceptions from it pleasing.

### DIFF
--- a/LiteDB/DbEngine/Disks/FileDiskService.cs
+++ b/LiteDB/DbEngine/Disks/FileDiskService.cs
@@ -268,6 +268,7 @@ namespace LiteDB
 
         private void TryRecovery()
         {
+            if (!_journalEnabled)return;
             // if I can open journal file, test FINISH_POSITION. If no journal, do not call action()
             this.OpenExclusiveFile(_journalFilename, (journal) =>
             {


mbdavid:
Hi @jaked122, even if you open database with no journal mode, your last use can be used and can be corrupted. To avoid this mscorelib exception isn't better just check before if file exists? Likte `if(!File.Exists(_journalFilename)) return;`

jaked122:
That'd also fix it, but I figured that checking a boolean would be quicker
than checking the file with every transaction. In my application, the most
reliable way I worked out for doing this had a great number of small
insertions.

I've unfortunately switched to mongodb because the datastore grew too large
with litedb, and the latency also grew too large for litedb.

Also, if you tell the database that it shouldn't journal, doesn't it make
more sense to not try and recover the previous transaction? As the
transaction should have never been recorded additionally in the first place.

mbdavid:
Yep, you right. Make more sense previus connection didn't use journal file (and yes, if much more fast your boolean check only).

PR #140, opened 2016-03-07
Removed dependency on Reflection.Emit

lee-orr:
I removed Reflection.Emit and replaced it with lambdas, to enable compilation & use in unity for webgl. Should also allow Unity & Xamarin iOS (untested)

### DIFF
--- a/LiteDB/Serializer/Mapper/Reflection.cs
+++ b/LiteDB/Serializer/Mapper/Reflection.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Reflection.Emit;
 
 namespace LiteDB
 {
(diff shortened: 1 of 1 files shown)


mbdavid:
Hi @lee-orr, nice job. Did you test performance about this? I use Emit because it's fast but require full .NET. I thinking in use compile condition to create 2 version: Full .NET and PCL.

lee-orr:
I haven't had a chance to test performance. It seemed to respond well for my use case, but it's a very limited case.

negue:
Is it possible to extract the Reflection.Emit calls (with an Interface) so that we could enable the use in the full .Net and all others could use the lambda version?

gentledepp:
I think negues idea is very good - if doable.
The reason being that Reflection.Emit is surely much faster, but on the other hand, you cannot use it on platforms such as Xamarin.Android and Xamarin.iOS

ToniMontana:
I have tested this commit on Xamarin.iOS and successfully tested LiteDB in a Xamarin.iOS application. It seems to work. A compile flag to be able to build for different platforms would be very useful.

mbdavid:
There is a new version in master branch that supports Xamarin.iOS. Can someonde test (I have no Xamarin skill)?

negue:
Can be closed now. :)

PR #176, opened 2016-05-31
Set values into collections with private setter

xszaboj:
This code checks if the original collection has no setter and if that is the case tries to set the deserialized collection into the original one.

Tests included

### DIFF
--- a/LiteDB.Tests/LiteDB.Tests.csproj
+++ b/LiteDB.Tests/LiteDB.Tests.csproj
@@ -88,6 +88,7 @@
     <Compile Include="Bson\JsonTest.cs" />
     <Compile Include="Index\LinqTest.cs" />
     <Compile Include="Mapping\IncludeTest.cs" />
+    <Compile Include="Mapping\UnitTest1.cs" />
     <Compile Include="Shell\ShellTest.cs" />
     <Compile Include="Utils\Initialize.cs" />
     <Compile Include="Utils\DB.cs" />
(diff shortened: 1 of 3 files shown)


mbdavid:
Hi @xszaboj, did you test for e `get` only property. In this case, LiteDB can't serialize/deserialize property

xszaboj:
I'll write more tests with read only collections.

xszaboj:
Hi David,
I run into a problem, but I don't know how to address it.
Maybe you going to have an idea.
I tried to write a new set of tests for read only collections.
So as long as it is a read only collection property without setter it works fine for me.
Problem is when the collection has setter like this:
public ReadOnlyCollection<int> ReadOnlyInWrapperWithSetter { get; set; }

I don't even know if it make sense to do this? However it is possible.
And deserialization of this property fails even in master branch of your version.
Because you are not able to create an instance of ReadOnlyCollection without parameter. (make sense).

I'll commit this failing test into my repository so you can have a look on it.

xszaboj:
Hi @mbdavid, I agree exceptions must be handled. Have you tried to run those test I've added?
The bug I found is in master branch too.

Also could you give me an example what you mean by this comment:
Hi @xszaboj, did you test for e get only property. In this case, LiteDB can't serialize/deserialize property
?

Thank you for your time ;) 
I would really like to add this functionality into LiteDB so people could use collection without setter as in EF.

mbdavid:
Hi @xszaboj, I'm backing to work in LiteDB and implement some features from users. This section (Mapper user class to BsonDocument) is quite complex because have a lot of options. Maybe I can put this in v2 version (or v2.1).
Can you check `master` branch and create another pull request? Now LiteDB support portable version so I need test in all platforms.

PR #180, opened 2016-06-12
fix for DeserializeList

sclif:
Happens when Enumerable was provided as an Object property during serialization. On DeserializeList it cannot find any Generic types or Interfaces as `type` is `System.Object`. Also, `type.IsArray` of Object is false even if underlying type is actually an array, which leads to a different can of warms. See the tests and comments.

I updated tests as well, but because it is not possible to detect what was the original underlying type it is hard to Assert expected and actual types. Therefore, `(ICollection)` cast.

Could be a wrong using of the LiteDB altogether. Stumbled upon this one when tried to serialize-deserialize `Dictionary<string,object>` where one of objects was `string[]`.


sclif:
It is no more relevant to me as I found another way how to achieve what I need. Feel free to close it if you think it is just wrong.

PR #183, opened 2016-06-27
Fixed queries returning incorrect results.

izackp:
Specifically for queries involving doubles. Not sure whats up with the newline thing (I'm using githubs interface), but I'm too lazy to fix it.


mbdavid:
Thanks @izackp !

PR #194, opened 2016-07-11
[1.0.4] Fix invalid start position in RecoveryService

nerai:
Start position was multiplied by page size even though it already scaled. This caused the target page position to be calculated wrongly, leading to a) database corruption or b) an exception due to int overflow.

Note that this change refers to the version tagged 1.0.4, *not* the current master. A quick glance at the current version suggested that this file was completely rewritten, so I'm unsure if it can still be applied. (Sadly, github did not let me create a pull request against the tag, only against a current branch.)

fix mbdavid/litedb/#193

### DIFF
--- a/LiteDB/Storage/Services/RecoveryService.cs
+++ b/LiteDB/Storage/Services/RecoveryService.cs
@@ -71,7 +71,7 @@ namespace LiteDB
         private BasePage ReadPageJournal(BinaryReader reader)
         {
             var stream = reader.BaseStream;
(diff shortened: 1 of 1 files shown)


mbdavid:
Hi @nerai, can you request this in `v1-fixes` branch? I can't change from here.

nerai:
Hi, I've just checked and the new v1 branch is already fixed, very good. Sorry I was a bit too late with the new pull request

PR #195, opened 2016-07-11
Various small fixes after merge

szurgot:
After you merged in my changes, I hit some small issues that turned into larger ones. I fixed up some of the failing tests, and got things working with my two test projects again. Here are the results of those fixes.

I hope the transaction changes aren't a problem, it helped me track down some issues converting to objects, and it made sense to leave them that way.

Reworks all transactions to be a disposable object so they can be wrapped in using blocks as well as tracked in transaction service
Updated all internal transactions to use Transaction object
Added try/finally to Find in LiteCollection to close the enumerator and finalize the transaction
Fixed VisitValue (QueryVisitor) to handle enums properly on all platforms
Accounted for issue where an index was added implicitly based on a BsonDocument would not crash (Reflection.cs)

### DIFF
--- a/LiteDB.NetCore.Tests/LiteDB.NetCore.Tests.csproj
+++ b/LiteDB.NetCore.Tests/LiteDB.NetCore.Tests.csproj
@@ -68,12 +68,6 @@
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\LiteDB.NetCore\LiteDB.NetCore.csproj">
-      <Project>{28c0558c-13b2-4f3d-b956-5d6081625f61}</Project>
-      <Name>LiteDB.NetCore</Name>
-    </ProjectReference>
-  </ItemGroup>
-  <ItemGroup>
     <Folder Include="Properties\" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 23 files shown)


negue:
Please wait with the merge, so that I can finalize my pullrequest and don't want to start over a 3rd time ^^

I can copy those fixes too, cause I changed many of the tests anyway, if its ok with @mbdavid

mbdavid:
Hi @negue, no problems. I will wait. @szurgot implmement a diferent (and nice) option with Transaction. I amazing how many minds think better and one.
I saw your source @negue, and you organize solution with a lot of plataform names (like Android and iOS). I have no knowledge about .net plataform (other than full .net) so I don't understand what the diferente. On big diference I saw is that you implement custom Reflection function (removing Emit) like this PR: https://github.com/mbdavid/LiteDB/pull/140 but @szurgot still with Emit. Is that right?

negue:
That source you saw was the one before the dev got merged to master. 

Emit is still working on .Net35 / .Net45 Platforms, on others I use Expression.Compile instead of Reflection (GetConstructor).

I splitted them up so that each Platform has their own platform specific calls, anyway once im Home I can push the final changes with a summary of what changed / removed.

Also nearly all tests working on for every platform, incl. encryption.

negue:
@mbdavid we should create a slack or gitter  group , easier to discuss ^^

szurgot:
@negue I see what you did in your previous commit, and it looks pretty good. If you're doing the same with the new commit, I can apply my changes on top of that pretty easily if necessary.

@mbdavid I did my changes in place because I didn't want to disrupt the existing project too much, but these changes will simplify the code base back down, and provide good cross platform support easily.

PR #196, opened 2016-07-11
LiteDB goes Cross-Platform

negue:
There are probably many debug steps, which I'll remove later, just wanted to push my current version of it.

I removed the .NetCore Project (which was NetCore, UWP and .Net46), because the .Net45 Library can be also used in .Net46, UWP has its own Platform. And for NetCore, I tried to create a simple website and using any sample pcl project..... but after 5 hours of wasted time, I kinda decided to wait to support NetCore until its easier to understand what and why it really works. ( Need help :stuck_out_tongue_winking_eye:  ) 

**Changes from master:**

- Crossplatform with all Tests!
- Shell is an own Class / Library (in PCL)
- FileDiskService and EncryptedDiskService merged together
- Encryption fully working on each platform! (also compatible encryption settings)
- LiteDatabaseFactory inorder to get an instance of ILiteDatabase
- There is a LiteDBPlatform / TestPlatform for each Target (which handles the specific stuff)

**Breaking changes:**
- need to have one call which sets the LiteDBPlatform.Instance
- more use of the LiteDatabaseFactory (if this is too much of a break, I'll remove it again and let the Databases be created "the old way")

**FYI:**
- #140 can be closed, cause Reflection.Emit is still be used on .Net35 / .Net45 (Desktop), Expression.Compile on all other PCL's
- #195 still needs to be applied after merge
- I readded some tests , in-order to see if all works as expected on that many platforms, so we maybe even need more tests
- the ConcurrentTest still not working, I don't know if its due my changes, maybe I forgot something from the master
- for the Tests on Mac I had to create a simple UI cause there isn't any test project available :sweat_smile: 

**Let the review begin!**

Btw, we should also use Gitter / Slack to discuss more.


mbdavid:
Hi @negue and @szurgot . I'm think how can we talk about this changes to cross plataform. This change are much more deep than @szurgot propose. 

@negue, I do a quick check in your source and have some questions:

- Are necessary LiteDatabaseFactory? Old way isn´t simple? I dont see devs using another implementation of ILiteDatabase than LiteDatabase or I'm not understand something...
- You exposes some classes and create interfaces like IIndexService and DbEngine (class) that was internal only.
- Did you remove shell from LiteDatabase class (Run method) but keeps LiteShell class inside LiteDB project?
- I guess that you "copy" some files instead "link as reference", like in "LiteDB.Core\DbEngine\DbEngine"

I think that the easy way to create this cross plataform is have only 1 project with real files (using #if condition) and all others vs project are a linked sources.

szurgot:
@mbdavid A chat room could help this go faster (and probably not distract others). I set one up on my repo and sent the two of you an invite.

After working with my implementation, I have found there is room for improvement there... Let's discuss over in the gitter chat (I sent the invite to both of you)

negue:
1. Removed the LiteDatabaseFactory, it was before I merged those DiskServices into one
2. & 3. in LiteDB-Project the Shell is still in the same Project, in PCL its separated to a different one, cause not everbody needs to use the Shell. Also the Shell needs to have access to the DbEngine, thats why its public and not internal (due that the Shell is now in an own Project) 
-> I could try to use "InternalsVisibleTo" if you want @mbdavid 
4. Good catch, I wanted to have all files linked , fixed that in a commit

PR #198, opened 2016-07-12
Lib update

szurgot:
Update after @negue merge with my additional changes.

Reworks all transactions to be a disposable object so they can be wrapped in using blocks as well as tracked in transaction service
Updated all internal transactions to use Transaction object
Added try/finally to Find in LiteCollection to close the enumerator and finalize the transaction
Fixed VisitValue (QueryVisitor) to handle enums properly on all platforms
Accounted for issue where an index was added implicitly based on a BsonDocument would not crash (Reflection.cs)


negue:
:+1:

- QueryVisitor (with enum) was already working on all platforms in my pull request, or did I miss something? Just changed the Type.IsEnum to Type.GetTypeInfo().IsEnum

szurgot:
That was an errant comment from my previous commit. (QueryVistor) Sorry

PR #208, opened 2016-07-17
Added check if Journal is enabled ( for 1.0.4 )

microvb:
This corrects the following problem :

When Journal is not enabled, TryRecovery() still attempts to access the journal file. If the journal file does not exist, this throws a FileNotFound exception which shows up in the debug output.


mbdavid:
Hi @microvb, can you request this to `v1-fixes` branch? I can change from here.

microvb:
See [https://github.com/mbdavid/LiteDB/pull/215](url)

PR #211, opened 2016-07-20
fix: tests, use temp-path in mobile tests

negue:
sorry about the spacings, its Xamarin Studio, never use it that much


mbdavid:
No problems! Thanks

PR #238, opened 2016-08-01
Adds a FileHandler specifically for UWP that uses System.IO instead o…

szurgot:
…f StorageFile to improve file access speeds.


LiteDB/Platform/LitePlatformWindowsStore.cs
```diff
@@ -28,7 +28,11 @@ public class LitePlatformWindowsStore : ILitePlatform
         // Making this private for now, because putting the folder anywhere but in the application store causes performance issues.
         private LitePlatformWindowsStore(StorageFolder folder, Func<string, IEncryption> encryption = null)
         {
-            _fileHandler = new LazyLoad<IFileHandler>(() => new FileHandlerWindowsStore(folder));
+#if WINDOWS_UWP
+            _fileHandler = new LazyLoad<IFileHandler>(() => new FileHandlerUWP(Windows.Storage.ApplicationData.Current.LocalFolder));
```
negue: you should use the parameter `folder`

szurgot: I didnt mean to include that signature in the push request. Just the body. Ill update when i get home.

Issue #259, opened 2016-08-08
Unexpected exceptions raised when working with a protected database but without password


falahati:
Well this is more or less a bug, but can be categorized at a feature request as well.
Problem is that we can detect if the provided password is incorrect for a database, but there is no way to detect if a database is protected in first place or not.

I strongly believe that the `LiteDatabase` constructor or its methods (or at least header loader or page reader) (in case of lazy loading) should raise a `LiteException` with `LiteException.DATABASE_WRONG_PASSWORD` as the exception code.

Currently, if you open a protected database with no password, all you get is a database that seems OK but may or may not raise unrelated exception (like not implemented exception, Bad pageType exception, etc) in unrelated and unexpected places.

mbdavid:
Uhm... I will do a check in DiskFile in datafile has password... so I can throw "Wrong password"

falahati:
Can be closed now then,

PR #261, opened 2016-08-08
Logic for checking password moved to the FileDiskService base class

falahati:
Addressing this issue: #259


negue:
@mbdavid I think `EncryptedDiskService` can be removed anyway, cause it's not being used anymore and was merged together to `FileDiskService`

falahati:
All encryption and decryption still goes through this class. Still, it can be merged.

But personally, I believe that we should keep it separate, in case of custom encryption methods.

In-fact, I believe that @mbdavid should make some of the `internal` classes `public` so developers can write their own custom code for things like this.

negue:
Just saw that my merge of `EncryptedDiskService` and `FileDiskService` got reverted again ... so ignore my last comment :/

Issue #260, opened 2016-08-08
Better string manipulation by ConnectionString class


falahati:
4th issue so far I believe in same day :)

Any way, currently ConnectionString class splits the passed string using the '=' and ';' characters which mean that this two characters can not be used as the password of database or even in the path of the database file. But then, both '=' and ';' are valid path characters in Windows. In fact, you cant be sure which character is valid and which one is not a valid character when writing for multiple platforms. I strongly ask for a better manipulation of connection string in ConnectionString class. At least we need a way to exclude any string inside two ' " ' characters.

Above problem is about this line:
https://github.com/mbdavid/LiteDB/blob/master/LiteDB/Utils/ConnectionString.cs#L23

And specially this one:
https://github.com/mbdavid/LiteDB/blob/master/LiteDB/Utils/ConnectionString.cs#L21

mbdavid:
Hi @falahati, yes, you right. I will fix this using regex, thanks.

falahati:
Solved. Thanks

PR #262, opened 2016-08-08
ConnectionString class now uses regex to manipulate string

falahati:
Addressing #260


falahati:
hmm,
Test failed when opening a file that doesn't exist; still I can't understand the reason this is related to my changes. File name seems alright any way and other tests get passes with same code. wired.

falahati:
Well I am going to close this one, till I found the reason for its failure.

PR #265, opened 2016-08-08
ConnectionString class now uses regex to manipulate string

falahati:
Well all the tests passed on my machine, so I am creating this request again.

Addressing #260


falahati:
well, problem solved. please take a look @mbdavid

PR #293, opened 2016-09-01
Fixed timespan parsing in ConnectionString

d-led:
Hi, I've been toying around LiteDB. Looks great! I've tried using the `timeout` connection string parameter, and it failed at type conversion. Here's a fix based on the [SO answer](http://stackoverflow.com/a/11967393/847349) by @mgravell.

Tested it on .Net 4 only. Hope it works on all platforms.


mbdavid:
Hi @d-led, I will merge this in a few days, ok? I will do some changes in ConnectionString parser to support values with "

d-led:
Hi @mbdavid ok!

PR #303, opened 2016-09-17
.netstandard 1.4 port

Davidblkx:
a quick port to support ASP.NET Core and other netstandard compatible projects,
i've used .Net45 version as reference except for the Rijndael implementation.
since most of the code is the same i've only port part of the tests


mbdavid:
Hi @Davidblkx, thanks for this! I'm merging in master and I will made some test to submit to nuget.

PR #306, opened 2016-09-20
New Connection String Parsing

falahati:
Regex replaced by manual enumeration of connection string characters for more flexibility

### DIFF
--- a/LiteDB/Utils/ConnectionString.cs
+++ b/LiteDB/Utils/ConnectionString.cs
@@ -13,60 +13,78 @@ namespace LiteDB
     {
         private Dictionary<string, string> _values;
 
-        private static readonly Regex keyValuePairRegex = new Regex(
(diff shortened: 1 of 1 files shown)


falahati:
This commit needs your work to solve conflicts with older version. Sry about that, this happend cuz of reverting the changes and me being lazy and not creating a new branch. :)

mbdavid:
StringScanner is present in LiteDB main lib and another copy in LiteDB shell. But ok, it's beter not use because next version will not support shell inside main lib (only in shell tool).

falahati:
Well, I need to fix a problem with addresses having space in it and I really need this changes to be able to do so. Can we expect this changes soon? :)

mbdavid:
Hi @falahati, yes. I will release this weekend 2.0.3 fixing minor problems

Issue #320, opened 2016-10-17
stuct is not deserialized correctly


CaptBrick:
All properties of deserialized struct are null/default. The following code reproduces the issue.

```
namespace LiteDB.Tests.Tests
{
    public struct StructValue
    {
        public string Field { get; set; }
    }

    public class Container
    {
        public Guid Id { get; set; }

        public StructValue Struct { get; set; }
    }

    [TestClass]
    public class StructTest : TestBase
    {
        [TestMethod]
        public void Struct_Test()
        {
            var m = new MemoryStream();

            using (var db = new LiteDatabase(m))
            {
                var col = db.GetCollection<Container>("col1");

                var id = Guid.NewGuid();
                col.Insert(new Container() { Id = id, Struct = new StructValue() { Field = "FieldValue" } });
                var item = col.FindById(id);
                Assert.AreEqual("FieldValue", item.Struct.Field);
            }
        }
    }
}
```

mbdavid:
Hi @CaptBrick, current version of LiteDB offers 2 ways to how create instances/get/set values. One is using Emit class (which are very fast) and one using expressions (which if more flexible). For current version 2, I will still using those both version just keep compatible.

But, I'm working on new version (v3). This new version will works only with expression to be same class used in .NET 3.5 and Portable/NetCore version.

CaptBrick:
@mbdavid As a workaround in my project I've registered custom Serializer/Deserializer to global BsonMapper. Is there a better way to to this? How can I switch to expression get/set for a single type? Also I don't see a v3 branch on github. Is v3 the current dev?

mbdavid:
Yes, v3 is dev branch. But, for now, it´s only avaiable LiteEngine, a new lower data access layer. This layer does not works with POCO classes (only BsonDocument).
For now, the better solution is use custom Serializer/Deserializer functions. Change Emit to Expression reflection needs create a new LitePlatform (works only with portable version).

mbdavid:
Now works in v3-beta. If you found any problem, please re-open this issue.

0xFireball:
Guid and DateTime are still zeroed when objects are loaded from database:
Code:
```
public class UserSession : DatabaseObject
    {
        [JsonProperty("id")]
        public string SessionId { get; set; } = Guid.NewGuid().ToString("N");

        [JsonProperty("userAgent")]
        public string UserAgent { get; set; }

        [JsonProperty("startTime")]
        public DateTime StartTime { get; set; }
    }
[...]

PR #321, opened 2016-10-17
#320  use property setter/getter invocation instead of il.Emit to fix…

CaptBrick:
I know this will most likely be rejected but I wanted to ask you why do you need to use DynamicMethod instead of simple Getter/Setter invoke? This also fixes the #320


mcmonkey4eva:
ILEmit is theoretically a LOT faster than reflection... that's why it's used.

mbdavid:
And, in my tests, Expressions are fast as Emit. Current version uses Expressions (and a little of Emit to specific case)

PR #324, opened 2016-10-19
Add punctuation in error messages; slightly improve a few message texts

nerai:
This is just a tiny cosmetic change. In accordance with the [suggestions of MSDN](https://msdn.microsoft.com/en-us/library/seyhszts%28v=vs.110%29.aspx) [1], added punctuation to error messages. Also improved a few messages slightly to make them easier to understand.

[1]
> Use grammatically correct error messages, including ending punctuation. Each sentence in a description string of an exception should end in a period. For example, "The log table has overflowed.” would be an appropriate description string.


mbdavid:
Hi @nerai, thansk. Could you PR to dev branch?

nerai:
I rebased it on dev.

Issue #325, opened 2016-10-22
Dependency Injection support?


raizam:
Hi,

I'm just getting started, and I have a question about the POCO, which litedb requires a public parameterless constructor.
Would it be possible to implement a IoC/dependency injection mechanism, where I can pass a Func<T> to liteDb as a instanciation resolver?

If not possible then I'll be looking in the source where POCO are instanciated on deserialization, and I'll implement it :)

Thanks

Edit: so I've found where objects are instanciated, at [BsonMapper.Deserialize.cs L155](https://github.com/mbdavid/LiteDB/blob/d6e920efd4d5bd0e76613e1bfd98370b211777d6/LiteDB/Serializer/Mapper/BsonMapper.Deserialize.cs#L155)
I know custom serializers are availlable, but I want to use the  property mapper and at the same time resolve types with parameterized ctors

mbdavid:
Hi @raizam, thanks for your PR, I will take a look closer. But for now, I'm working in `dev` branch for next release - this next release will change a lot from now. My plans is keep v2 as is, works only in bugfix.

Is `Emit` reflection supported in NETStandatd?

raizam:
Hi @mbdavid, I do not know much about .NETStandard, but I'll take some time too see if the code I added works as expected.

If everything runs fine, I don't see how these addition could affect anyone else, I believe it should be fine.
The IoC support don't change any behavior, and no Api break.

The Struct property support might affect users that used public structs properties as transient data. But it should be uncommon, and I'm not sure if even it's the case, it would affect that much. But still this is indeed a behavior change, even is it's very minor. Anyway I have my own branch, so I'm fine if it's not merged :)

What will be the main changes in the next version you are working on?

mbdavid:
Hi @raizam, the main change in next version is transforma LiteDB in a thread safe database. Also, LiteDB will support multi key index (index all values from an array document field). All this changes (an others more) are on database engine.

In portable area, my idea is convert LiteDB to NetStandard to avoid compile/initialize to many platforms. Next version will have only 2 version: .NET 3.5 and .NETStandard 1.3 compatible.

Now, I'm working on this upper layer: converting POCO entities to BsonDocument and Linq operations. Thats why this PR is important. I will merge them to my dev branch

mbdavid:
Now I was checking your customTypeInstanciator. Isn't better to set new instance function only when nedded? Some like this:

```
BsonMapper.Global.Entity<Customer>((value) => new Customer())
[...]

PR #326, opened 2016-10-22
 IoC support + Struct Property support

raizam:
Minimal changes to BsonMapper's constructor, now it takes an optionnal
Func<Type,object> parameter, to let the user specify a custom type instanciation function.
implements #325


raizam:
Well I needed structure support as well ( #320 ), so I added it, still using Emit

PR #331, opened 2016-11-01
include inner exception information (if available) when construction fails

nerai:
In reflection.cs, there is information lost when construction fails. This PR includes the thrown exceptions, if any, as inner exception, to make debugging these situations easier.


mbdavid:
Thanks @nerai.

PR #357, opened 2016-12-03
Change transaction syntax to make Commit() implicit

apkd:
The current transaction syntax is the following:

```C#
using(var transaction = db.BeginTrans())
{
    // (operations)
    transaction.Commit();
}
```

This is unclear and prone to mistakes. Forgetting to commit the transaction will roll it back, or in certain cases (nested transactions) lead to difficult to diagnose bugs. For example, modifying the sample from the wiki:

```C#
using (var transaction1 = db.BeginTrans())
{
    col.Insert(new Person { Id = 1, Fullname = "John" });

    using (var transaction2 = db.BeginTrans())
    {
        col.Insert(new Person { Id = 2, Fullname = "Joana" });
        //transaction2.Commit(); // we don't commit the inner transaction
    }

    col.Insert(new Person { Id = 1, Fullname = "Rick" });
    transaction1.Commit();
}
// what's the result? surprise! only the third person (Rick) gets comitted.
```

I propose a new syntax that makes the `Commit()` statement implicit at the end of the transaction.
```C#
using(var transaction = db.BeginTrans())
{
    // (operations, no need to commit)
}
```

The `IDisposable.Dispose()` implementation uses `Marshal.GetExceptionCode()` to decide whether the transaction should be committed or rolled back. I preserved the behaviour of `Rollback()` only cancelling the operations until the rollback statement, not withing the scope of the entire transaction.

I am not certain what behaviour is intended in some cases so please review the code and decide whether this kind of solution is acceptable.

I am also adding several tests for the transaction system.

### DIFF
--- a/LiteDB.Tests/Database/TransactionTest.cs
+++ b/LiteDB.Tests/Database/TransactionTest.cs
@@ -1,0 +1,107 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace LiteDB.Tests
+{
+    [TestClass]
+    public class TransactionTest
+    {
+        [TestMethod]
+        public void TransactionCommit_Test()
+        {
+            using (var f = new TempFile())
+            using (var db = new LiteDatabase(f.Filename))
+            {
+                var col = db.GetCollection<Person>("Person");
+
+                col.Insert(new Person { Fullname = "John" });
+                col.Insert(new Person { Fullname = "Doe" });
+                using (var transaction = db.BeginTrans())
+                {
+                    col.Insert(new Person { Fullname = "Joana" });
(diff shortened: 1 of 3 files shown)


mdissel:
Have you ever worked with ado.net transactions? the default behavior is to rollback.   See https://msdn.microsoft.com/en-us/library/2k2hy99x(v=vs.110).aspx

mbdavid:
Hi @apkd, thanks for your contribution. To be honest with you, I prefer your way to implement transaction (and it's new to me that you can get exception inside Dispose). To me, less code is better :). But, I posted in gitter about a month ago how was better way to implement transaction and write this samples: https://gist.github.com/mbdavid/9a8daddb1fe8e73732cf156f5b622022

The way was implemented is more "standard" and it's used in many libraries (including ado.net, TransactionScope and EF).

PR #382, opened 2016-12-16
Fix for NullReferenceException when deserializing a deleted DbRef

andypelzer:
Hi,

I encountered a NullReferenceException during deserialization. The PR includes a test that reproduces the exact scenario. A quick summary of the steps:
- create and insert an entity that has a list, which is a DbRef
- create an instance of the referenced entity, add it to the first entity and store both
- delete the referenced entity (directly, not by removing it from the list of the first entity)
- load the first entity again => throws

The fix results in ignoring the deleted entity and just returning an empty list.

### DIFF
--- a/LiteDB.Tests/Database/NullRefTests.cs
+++ b/LiteDB.Tests/Database/NullRefTests.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


mbdavid:
Thanks @andypelzer, I will merge in some days after I merge dev branch to master.

andypelzer:
Thank you!

PR #403, opened 2016-12-27
Connection string builder

mstum:
This is useful because it makes it easier to create a correct Connection String for LiteDB (e.g., no worrying about quoting/escaping parameters - in fact, no roundtrip through a string at all if the new LiteDatabase(ConnectionString) ctor is used).

My use case: I have a desktop app where the user uses a File->Open menu and enters a password. Since I have these two pieces of data as separate strings in my application, this is more robust than having to create a string with the correct syntax (and escaping in case the password contains special characters).

### DIFF
--- a/LiteDB/Database/LiteDatabase.cs
+++ b/LiteDB/Database/LiteDatabase.cs
@@ -36,12 +36,13 @@ namespace LiteDB
 
         #region Ctor
 
-        /// <summary>
-        /// Starts LiteDB database using a connection string for file system database
-        /// </summary>
-        public LiteDatabase(string connectionString, BsonMapper mapper = null)
+        public LiteDatabase(ConnectionString connectionString, BsonMapper mapper = null)
         {
-            _connectionString = new ConnectionString(connectionString);
+            if (connectionString == null)
+            {
+                throw new ArgumentNullException("connectionString");
+            }
+            _connectionString = connectionString;
             _log.Level = _connectionString.Log;
 
             if (_connectionString.Upgrade)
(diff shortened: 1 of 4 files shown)


mbdavid:
Hi, after a long time after.... sorry for this long delay. Let me ask why create new class builder to write an string connection if we can use ConnectionString direct class(ok, not now because has only string ctor initializer and has no ConnectionString in LiteDatabase ctor)?

mstum:
Hi, this was initially mirroring the idea [SqlConnectionStringBuilder](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlconnectionstringbuilder(v=vs.110).aspx): Have 1 Factory that I can configure and then create an immutable ConnectionString that I can pass around. It doesn't have to be it's own class, especially since `ConnectionString` already exists as a class (instead of just a string).

The idea behind having a builder at all is to not have to worry about properly escaping parameters when building a string (e.g. worrying about spaces in the filename or whether it needs to be quoted or not).

So putting it into the existing `ConnectionString` class works, I just wonder if there is an issue if I have created a `ConnectionString`-object and store it (e.g., for Dependency Injection) and then need to create a new one by just modifying e.g., the Filename or Password of the instance that I have. Now I've changed the filename for all objects that reference that one ConnectionString. By having a separate Factory and an immutable ConnectionString, it is clear to the developer that I have to make sure that all objects that need the new ConnectionString get it, but the risk of me accidentally corrupting data because I didn't realize I was modifying a shared reference is lower.

But I don't think that's a problem, Developers should know to create a new ConnectionString object if they need to create a different ConnectionString.

mbdavid:
Hi @mstum, I implemented ConnectionString with get/set properties. I undersant your worry about dependencis, but connection string is already a builder class to LiteDatabase ctor with all paramters. As you can see here https://github.com/mbdavid/LiteDB/blob/master/LiteDB/Database/LiteDatabase.cs#L50 ConnectionString object "dies" inside this ctor, because all options are copied to another objects.

BTW, thanks again for contrib 👍

Issue #472, opened 2017-02-09
Problem using auto generated service reference objects - vol2


MiklosPathy:
Hi there,

A very similar problem returned with 3.0.0 stable, what is described in #340 .
The problem is the same: the BsonMapper and the serializator serializes fields not meant to be serialized. Of course, the deserializator dies on them.
In the 3.0.0 betas this problem was not there, it reapperared only in V3.0.0 stable.

My workaround for this is the same as I recommended in the #340 for V2.0.x branch: simply ignore the ignorable fields.

I can not make a pull request, because i dont have the necessary knowledge, time and tools here now so i simply write here what should be modified:

Insert this method into the BsonMapper.cs 

```
        /// <summary>
        /// Override this method to fine tune what should fields should be mapped and what not.
        /// By default it ignores only the fields with BsonIgnoreAttribute.
        /// </summary>
        protected virtual bool IsIgnorableMember(MemberInfo memberInfo)
        {
            var ignoreAttr = typeof(BsonIgnoreAttribute);
            return memberInfo.IsDefined(ignoreAttr, true);
        }
```
Remove the 
`            var ignoreAttr = typeof(BsonIgnoreAttribute);`
line from the BsonMapper's BuildEntityMapper method.

Replace the
`            if (memberInfo.IsDefined(ignoreAttr, true)) continue;`
line in the BsonMapper's BuildEntityMapper method with this
`            if (IsIgnorableMember(memberInfo)) continue;`

With this solution it is possible to override what should be ignored by the Mapper / Serializator.

I hope this modification can be done quickly, by someone in the main branch. 
Meanwhile I am strarting to learn how to use GitHub.

Happiness!

mbdavid:
Hi @MiklosPathy, I was reviewing source and has amost no diference in BsonMapper from beta3 to final release. Take a look here:

https://github.com/mbdavid/LiteDB/compare/v3.0.0-beta3...master#diff-4727e7744b7e8b1bdaeb1c5e09950d9c

But, I remeber you when I was writing new BsonMapper. So, it's possible to write a func to made your test (in current version):

```
// assuming that you are using global instance:
BsonMapper.Global.ResolveMember = (type, memberInfo, member) =>
{
    // type is Type of your class
    // memberInfo is your property/field
    // member is LiteDB.MemberMapper (that maps your poco class/properties to BsonDocument)

    // if you want ignore some property, just set member.Name == null
    if (memberInfo.Name == "ExtensionData")
    { 
[...]

PR #473, opened 2017-02-09
Chance to override field ignore in BsonMapper

MiklosPathy:
The code for the problem described in issue #472


MiklosPathy:
Solved with recommended solution.

PR #488, opened 2017-02-15
Fix CreateInstance in multithreading scenario

masterdidoo:


mbdavid:
Thanks, this second check avoid 2 threading try insert both in _cacheCtor.

Issue #502, opened 2017-02-24
LiteDB keeps the database open even after throwing an Exception


falahati:
Hi David,

While porting the LiteDBViewer to use the newly released LiteDBv3, I noticed that the LiteDB keeps the file open while throwing an exception for the version or invalid password.
As the result, I can't upgrade the file after getting an exception for invalid version.

Please take a look into this and close all file handles before throwing the exception.

mbdavid:
Hi @falahati, I don't get where my file can still be open. To upgrade, I use:

`using(var stream = new FileStream(filename, System.IO.FileMode.Open, FileAccess.Read)) { .. }`

Inside this block, I tests for version file and password. If any exception throwed, file must be closed when close `using` by `Dispose()`.

falahati:
Yeah, it should. But this is not what we are seeing as the end result.

Let me take a look at the source code, I will get to you with a pull request if I managed to find the problem.

mbdavid:
Ok @falahati, now I got it! Your PL make sense. If an throw occurs during ctor on `using` statment, there is not "dispose" method call. In this case, _disk will keep opened.

```
using(var db = new LiteEngine(.....))
{
    ...
}
```

If any error occurs inside `LiteDatabase(ctor)`, `Dispose` will not be called.

Thanks again!

falahati:
True,

`ctor` happens after creating the reference but before returning it. So at this point, `using` don't even have the reference to the `LiteEngine` to call its `Dispose` method.

Your comment made me take another look at the source code and now, I think that we may need to add another line to the `catch` part about `_crypto` object, like the one you have inside `Dispose` method. Or even better, instead of calling for `_disk.Dispose();`, call the `this.Dispose()` right there.

PR #505, opened 2017-02-25
Fixed a problem with not closing the file handle before throwing an exception

falahati:
Hey David,

Fixing #502


falahati:
Btw, All tests are passing on my machine.

PR #521, opened 2017-03-05
Makes id property be calculated deterministically

szurgot:
And also only allows one field to be considered _id.

All tests pass in both 3.5 and netstandard builds.

### DIFF
--- a/LiteDB.Tests/LiteDB.Tests.NetStandard.csproj
+++ b/LiteDB.Tests/LiteDB.Tests.NetStandard.csproj
@@ -22,7 +22,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>bin\Debug\netstandard\</OutputPath>
     <DefineConstants>TRACE;DEBUG</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
--- a/LiteDB/Mapper/BsonMapper.cs
+++ b/LiteDB/Mapper/BsonMapper.cs
@@ -309,9 +309,11 @@ namespace LiteDB
             var indexAttr = typeof(BsonIndexAttribute);
             var dbrefAttr = typeof(BsonRefAttribute);
 
-            var idProp = getIdProperty(type, idAttr);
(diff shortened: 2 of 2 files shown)


szurgot:
I looked at the test that failed in AppVeyor, and I don't think it has anything to do with this change. It appears to be happening because the file doesn't exist before the test starts, and when the two threads start, they are both creating the file. If one thread initializes the file fully, it works fine, but in a few instances if the file is "created" but the initial header isn't written in time, the other thread will read the header, which ends up being null, and then throw this exception.

I was able to get around this by adding a "count()" call to the table, which forces the creation before the two threads start adding records.

xied75:
@szurgot I got same failed test as the appveyor with a fresh cloned repo build. From what you described above, shouldn't we try to fix the root cause too? It's a bit scary to have these "random" failure. But most of the time it uncovers potential major bugs on threading/concurrency design.

@mbdavid ?

mbdavid:
Hi guys, about this appveyor fail, i test many times in my machine and works fine. All tests, running multiple times, always run ok. But I will take a look better on this, because it´s always same unit test method.

szurgot:
Obviously, it would be best to fix the problem, but I do think it's very much an edge case to think that you are going to have a program that will try to read from a brand new database on two (or more) threads simultaneously. More than likely, any program will have a little setup before it gets into multi-threading and that setup will do something to trigger the file to be created.

@mbdavid The other tests that do multi-threading already have a command that will create the file: db.EnsureIndex("col", "thread");

So, I think you've encountered this before.

szurgot:
I looked at the ID test a little closer, and there is a block at line 29 that looks like it should setup the database "var dbInit = new LiteDatabase(file.Filename)", but when the block is done, there is still no file. I think there should be something in that block like ensureIndex that will cause the file to be created, at which point all other multi-threading work will kick in.

szurgot:
Here's a better PR without the extraneous loop, and also removes a second call to getTypeMembers.

Also, for most cases, the time spent in BuildEntityMapper will be miniscule compared to some methods. Reflection itself is slow, but IO is invariable slower, and this method will be hit once for a given collection, but there could be hundreds of writes...

mbdavid:
Restored `GetId` from v2 with some modifications

https://github.com/mbdavid/LiteDB/commit/18941094cad49751cea33b4b4ea0b27bb8fd6237

Now supports Fields and NonPublic Id too (is setted in mapper)

PR #527, opened 2017-03-12
add osx for Travis build

xied75:
Now we also build for osx: https://travis-ci.org/xied75/LiteDB/builds/210166590

:)


mbdavid:
Thanks @xied75, thats helps a lot!

PR #556, opened 2017-03-29
Fix NullReferenceException when using FileDiskService during Shrink

mkosieradzki:
temp.FileLength will throw NullReferenceException after engine is Disposed, because LiteEngine.Dispose() disposes also DiskService (temp).


mbdavid:
Thanks @mkosieradzki!

PR #575, opened 2017-04-15
Lower min netstandard version

Bond-009:
The minimum net standard version was 1.4, now it is 1.3


Bond-009:
This isn't needed anymore

Issue #540, opened 2017-03-19
Unable to load LiteDB.dll using Powershell 5.1 on Windows 10


lboening:
I cloned the repository. I opened the solution in Visual Studio 2017 Community. I built the solution. Using Powershell, one is able to load assemblies (most of the time).

When loading:

Set-Location -Path "C:\code\LiteDB-3.1.0\LiteDB-3.1.0\LiteDB\bin\Release\netstandard"
Try {get-childitem"*.dll" | foreach { Write-host $_.Name; $assembly = Add-type -path $_.FullName}} catch { $error[0].Exception.LoaderExceptions}

I get this error:

`Could not load file or assembly 'System.IO.FileSystem.Primitives, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. 
The system cannot find the file specified.
Could not load file or assembly 'System.IO.FileSystem.Primitives, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. 
The system cannot find the file specified.`

lboening:
The following worked. I used Powershell 5.1 Package manager and found the compiled dll at Nuget.org.

`
find-package -name litedb | install-package -Credential (get-credential -Credential luke) -scope currentuser
Set-Location -Path 'C:\Users\Luke\AppData\Local\PackageManagement\NuGet\Packages\LiteDB.3.1.0\lib\net35'

Try {get-childitem"*.dll" | foreach { Write-host $_.Name; $assembly = Add-type -path $_.FullName}} catch { $error[0].Exception.LoaderExceptions}
`

lboening:
If this helps, I am making a bit of progress. I am able to use the LITEDB.DLL in a Visual Studio 2017 community project. I can create a DLL that I can get Powershell to load as an assembly. 

Next, I will see if "unblocking" helps. Thought I did that, however.

mbdavid:
Hi @lboening, sorry for not helping, I never used powershell and I dont know how works. But if it's works, let's share this in LiteDB wiki pages

wgross:
@lboening @mbdavid 
I think the reason of this is the global declaration of dependencies in the project.json. net461 doesn't need them and more importantly doesn't provide these assemblies and therefore loading fails. 

I've change the project.json (see pull request) and its working now with powershell on my machine.

lboening:
[...]

PR #576, opened 2017-04-16
Powershell Loading Issue/Dependencies in project.json

wgross:
I want to propose a solution of the powershell loading issue: #540 

Powershell is built with net461 but doesn't need the dependencies netstandard requires because the full framework provides all the classes

Moving the dependencies from a global declaration to a framework specific location solved the issue for me.  I can load LiteDB powershell again. 

I've also had to add a global.json because I',m using the latest tool chain.


lboening:
Yes, I would like to see if I can build on my machine. I am using Visual Studio Community 2017.

PR #577, opened 2017-04-16
BsonDocument copies its properties by reference  to KV array

### LINKED ISSUE #557: Missing Implementation of BsonDocument.CopyTo
Just stumbled upon this in BsonDocument class
 ```
public void CopyTo(KeyValuePair<string, BsonValue>[] array, int arrayIndex)
{
    throw new NotImplementedException();
}
```
Is this an unsolved problem or just a bug?
Because it also prohibits List.AddRange I was forced into a verbose workaround :-)

`
<bson document>.Aggregate(new List<KeyValuePair<string, BsonValue>>(), (l, kv) => { l.Add(kv); return l; });`

wgross:
Finally found some time to add code to copy a BsonDocument to an array.
This fixes #557 

I've also added a test because I'm a TDD guy. Please feel free to throw it away if you think this is over the top... i'm used to it :-)


mbdavid:
Thanks! Merged!

Issue #574, opened 2017-04-14
Power-off is not handled properly


mkosieradzki:
During tests I have found out that power-off events are not handled properly.

And my database is getting corrupted.

@mbdavid 
My **guess** is that you are not calling Flush/FlushAsync before closing the journal file. It basicaly means that JournalFile can be still in the OS Cache (and not on the physical disk yet) while you start to alter database file.

The same applies to the database file that should be flushed to the physical disk before cleaning up the journal file.

Closing the file does not mean it is physically flushed to disk.

mkosieradzki:
Of course I mean FileStream.Flush(true).

mkosieradzki:
@mbdavid Can you have a look, please?

mbdavid:
Sorry @mkosieradzki, I had almost no time for LiteDB for a long time. You code seams ok, I'm merging with master. Thansk!

mkosieradzki:
Thanks a lot @mbdavid! However, please remember this might have some performance impact, especially on non-SSD disks. In the future, I believe, that disk sync should be done asynchronously, but consistently journal+db.

mbdavid:
Yes, I know this. But it OS doesnt hang this pending saves this could be corrupt datafile. It;s on my plans to next major version be async read/writes, but I still don't know how do it (I never work with async in c#)

mkosieradzki:
I would be happy to help (I have a lot of experience with async-programming in c#) and if you are interested in getting some async prototype (for example for the DbEngine layer), I can try to hack something together in a separate branch for example in June/July.

mbdavid:
@mkosieradzki will be great to get this contrib! My idea for next version will be more simple & robust as possible, so I will try focus my dev only in LiteEngine and write more tests for better concurrency support.

mkosieradzki:
@mbdavid Good news is that I haven't noticed **any** performance degradation (on my SSD) after applying the FlushAsync patch! Just wanted to let you know :)

mbdavid:
Thats nice to know @mkosieradzki! Thanks

PR #582, opened 2017-04-20
Added flushing OS cache to medium

mkosieradzki:
This solves #574 . However only the for the portable version, because Flush(true) API has been added in .NET Framework 4 (and works properly since .NET Framework 4.5). So it will not work in .NET Framework 3.5.


lidanger:
How about Flush(true) in  .NET Framework 4?

lidanger:
Can you compile the code to other version of the .NET Framework and add to the Nuget? @mbdavid

mbdavid:
@lidanger, yes, both in .net35 and netstandard. But its implemented only for netstandard (so, will not work in .net 4)

```
#if !NET35
            _journal.Flush(true);
#endif
```

PR #627, opened 2017-06-08
throw meaningful exception instead of crashing with a nullreferencexc…

ghost:
When mapping a reference, the referee should have a BsonId-attribute defined; If not, this will result in a crash from a nullreference-exception.
It took some of my time to figure that out, so I suggest throwing an xception with a meaningful message so the developer knows better what might cause this exception.

### DIFF
--- a/LiteDB/Mapper/BsonMapper.cs
+++ b/LiteDB/Mapper/BsonMapper.cs
@@ -468,6 +468,11 @@ namespace LiteDB
             {
                 var idField = entity.Id;
 
+                if (idField == null)
+                {
+                    throw new LiteException(LiteException.ID_KEY_NOT_FOUND, "No id-field found for {0}. Make sure to set it for this type.", entity.ForType.Name);
+                }
+
                 var id = idField.Getter(obj);
 
                 return new BsonDocument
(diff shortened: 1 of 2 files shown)


NatriumDev:
Why does Travis fail?

mbdavid:
Hi @NatriumDev, I dont know. In my last commit same error occurs with same message (and I didn't change anything in Travis)

Program.cs(8,15): error CS0017: Program has more than one entry point defined. Compile with /main to specify the type that contains the entry point. [/home/travis/build/mbdavid/LiteDB/LiteDB.Tests/LiteDB.Tests.dotnet.csproj

DustinLuck:
This appears to be obsolete now: the missing-_id guard for the scalar DbRef path is already present in the current codebase in LiteDB/Client/Mapper/BsonMapper.cs at line [259](https://github.com/litedb-org/LiteDB/blob/c9d78eb900854309ee0756be056a2afbe1c6c24c/LiteDB/Client/Mapper/BsonMapper.cs#L259) (if (idField == null) ...). The change was introduced by commit [4785992](https://github.com/litedb-org/LiteDB/commit/47859926209da5aa1d5d4df61a72f0126af9a1ba) (“Add better error message in interface with dbref”).

I recommend closing this PR.

JKamsker:
Thanks for this! The same guard has been in the mapper since v5: RegisterDbRefItem now throws "There is no _id field mapped in your type" when the referenced type has no id, so I'm closing this as already fixed.

PR #632, opened 2017-06-13
Should fix travis build

RytisLT:
remove unused file that causes travis build to fail


RytisLT:
need to get my build tools right, I meant of course travis and not maven :)

RytisLT:
fixed commit message, all hail force push :)

mbdavid:
Travis are in a never-finish building...

RytisLT:
Removed the osx build from travis. It lookw like it timed out because unit tests take longer than 10 minutes to execute and during this time no output is written to console so travis assumes that something went wrong and kills the process.

PR #645, opened 2017-07-10
Remove unused BsonDocument.MAX_DOCUMENT_SIZE

olehfb:
The maximum document size of 1 megabyte is no longer enforced. Creating bson documents larger than 1 megabyte works fine. BsonDocument.MAX_DOCUMENT_SIZE is not referenced by any code. (The 512 byte size limit for indexed fields remains).

### DIFF
--- a/LiteDB/Document/BsonDocument.cs
+++ b/LiteDB/Document/BsonDocument.cs
@@ -7,8 +7,6 @@ namespace LiteDB
 {
     public class BsonDocument : BsonValue, IDictionary<string, BsonValue>
     {
-        public const int MAX_DOCUMENT_SIZE = 256 * BasePage.PAGE_AVAILABLE_BYTES; // limits in 1.044.224b max document size to avoid large documents, memory usage and slow performance
-
         public BsonDocument()
             : base(new Dictionary<string, BsonValue>())
         {


mbdavid:
Hi @olehfb, maybe I changed some code and forgot to add this validation. But you think it's better has no limit for each document? Documents are "in-memory" objects, must be readed full into memory and when need change, must serialize again and write all data into pages.

Mongodb limit in 16mb and DocumentDB 512kb

olehfb:
In my opinion it's best not to enforce limits which are not necessary.

For my use case for example, I will be storing documents that are way bigger than 1 megabyte. It's not a consumer facing application and will be running on a single high-end server with tons of RAM and multiple RAIDed SSDs. So I am not concerned about performance for my use case. First thing I did was tried to modify LiteDB source code to get rid of this 1mb limit, and that's when I found out it wasn't enforced any more.

I think it's best to warn developers about using large documents (maybe in the Wiki), but then let them make the final decision.

mbdavid:
Hi @olehfb, thanks! It's nice have some information about how others are using LiteDB... I will change documentation to just recommend write small documents if has many read/writes.

nick-goloborodko:
I like the idea of recomending the minimal possible document sizes but not enforcing it. I'm using LiteDB in a UWP applciation that captures and serializes data from the InkCanvas control and then storing it in LiteDB as a byte[] on the document. Each drawing serializes to about 15-20kb in size - but ocasionally there can be as many as 60-70 drawing in the same document - which would hit the 1Mb limit. The applciation does not do any heavy data processing in LiteDB - and since it is running on the client directly - theres only about 1 read and one write every 5 or so minutes,

PR #648, opened 2017-07-11
Fixed "Property was not mapped into BsonDocument" errors on Expressio…

pauldbau:
…n queries that involve enum properties

It seems that the latest Xamarin (which recently moved to Mono 5.0+ under the hood) compiles Expressions differently - in particular references to Enum properties are now wrapped in a Convert() call.

e.g. col.Find(t => t.MyEnumProp == MyEnum.Something) 
 - used to compile to a clean "t.MyEnumProp" on the left hand side
 - now compiles to "Convert(t.MyEnumProp, Int32)" in latest Xamarin

This code change caters for the potential Convert() wrapping and extracts the inner Operand in order to restore the expected behaviour of the GetPath() method, which in turn is used in various Mapper and Query areas of LiteDb.

### DIFF
--- a/LiteDB/Utils/Extensions/ExpressionExtensions.cs
+++ b/LiteDB/Utils/Extensions/ExpressionExtensions.cs
@@ -22,6 +22,14 @@ namespace LiteDB
             // quick and dirty solution to support x.Name.SubName
             // http://stackoverflow.com/questions/671968/retrieving-property-name-from-lambda-expression
 
+            // enum properties seem to get compiled with Convert(prop, Int32) wrapper calls on Mono 5.0+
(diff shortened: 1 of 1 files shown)


pauldbau:
Just wanted to also say that this is a great project and appreciate the tons of work that @mbdavid and contributors have put in.
We're using LiteDb in production (switched from using sqlite) and its been perfect for our needs.  Especially awesome since Android N screwed with Sqlite access, but also love avoiding the x86/64 hell that is trying to use sqlite on Windows.

mbdavid:
Thanks @pauldbau! I will merge in my current dev branch

Issue #652, opened 2017-07-18
LiteCollection.Find Contains works with Array but not for IEnumerable


RytisLT:
public class ItemWithEnumerable
    {
        public int[] Array { get; set; }
        public IEnumerable<int> Enumerable { get; set; }
    }

This works:

    LiteDatabase.Find(i => i.Array.Contains(1));

This does not:

    LiteDatabase.Find(i => i.Enumerable.Contains(1));

PascalSenn:
Do report a similar behavior for List<T>.

JensSchadron:
LINQ gets translated to the Query.Contains, but I noticed as well that it doesn't work with IEnumerables. Marked it as bug.

JKamsker:
This is fixed on the `dev` branch. `dev` ships as the LiteDB 6.x prereleases.

Please try the latest prerelease (currently **6.0.0-prerelease.81**):

```
dotnet add package LiteDB --prerelease
```

If the problem still happens on the prerelease, comment here with the version you used and a repro, and we'll reopen the issue.

PR #653, opened 2017-07-18
add support for Contains in IEnumerable #652

RytisLT:
LiteCollection.Find(i => i.List.Contains(1)) works with this commit


JKamsker:
Thanks for the fix and the tests! v5 rewrote the LINQ layer, and Contains on arrays, lists and any IEnumerable<T> property now translates to an ANY = expression (EnumerableResolver/ICollectionResolver), so this is already covered and I'm closing it.

PR #657, opened 2017-07-22
Added TFM for net45 to avoid improper referencing of net35 in net45+

### LINKED ISSUE #656: Problem with TargetPlatforms and the new .NET tooling
There are two issues here - first one let me discover the second one.
Current nuget package contains to target platforms:
netstandard1.4 and net35

1st issue: In package version 3.1.1.0 net35 assembly version is 3.1.0.0 when netstandard1.4 assembly version is 3.1.1.0.

And now this made me discover the second issue:
(at least when using the latest .net standard 2.0 tooling - preview 2) net461 seems to prefer net35 assembly version over netstandard1.3) this causes tools to use wrong version of assembly (net35 which has many many features disabled).

Solution:
1. Fix build to apply versioning corectly
2. Add net40 or net45 target with basically the same code as netstandard1.4
In that case when referencing LiteDB from net461 net40 will take precedence before net35 which is not a really good package.

UPDATE: Regarding the second point I am not really convinced whether it is true or not... But using net40/net45 tfm seems like a safe way.
UPDATE2: I am positive this is the issue. Please click following link and you will see that net35 wins.
http://nugettoolsdev.azurewebsites.net/4.2.0/get-nearest-framework?project=net461&package=net35%0D%0Anetstandard1.4
My recommendation will look as follows:
http://nugettoolsdev.azurewebsites.net/4.2.0/get-nearest-framework?project=net461&package=net35%0D%0Anet40%0D%0Anetstandard1.4

mkosieradzki:
Fixes #656 

I didn't bump the version.
Please note that I have changed the netstandard1.4 reference from Release\netstandard to Release\netstandard1.4 (generated by LiteDB.dotnet.csproj)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
   include:
     - os: linux
       dotnet: none
-      mono: 4.6
+      mono: latest
       dist: trusty
 before_install:
   - ./.travis/before-install-$TRAVIS_OS_NAME.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ language: csharp
 matrix:
   include:
     - os: linux
-      dotnet: none
+      dotnet: 1.0.3
       mono: latest
       dist: trusty
-before_install:
-  - ./.travis/before-install-$TRAVIS_OS_NAME.sh
+#before_install:
+#  - ./.travis/before-install-$TRAVIS_OS_NAME.sh
 
 script:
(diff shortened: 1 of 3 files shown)


mkosieradzki:
BTW. I have tested this and adding net35 target to this project file as well would allow to use new toolchain to build all versions: netstandard1.4, net45, net35.

We can also use the new toolchain to generate the nuget package instead of nuget.

PR #660, opened 2017-07-22
VS2017 upgrade

schulz3000:
#656
@mkosieradzki was working on a similar solution these days. To avoid having single solution files for every Framework Verison I unify everything in one Solution.

Also go down to netstandard1.3


mkosieradzki:
@schulz3000 
I have noticed that you have even added TFM for net451.

You can definitely take precedence before #657 and #659 ! I will close those PRs when your is merged.

BTW. You can probably even lower net451 to net45 AFAIR.

mkosieradzki:
Also I think it would be a good idea to remove AssemblyInfo.cs and use new csproj options for that purpose if @mbdavid agrees. Also I think it would be a great idea to replace nuspec with settings in csproj.

I would create a single version.targets with VersionNumber and reference it from all projects. It will be much easier to bump version numbers.

BTW. Fingers crossed for your fight with Travis :).

mbdavid:
Hi guys, I didn´t download this source yet but are you telling me that in VS2017 is possible not have multiple solution for multiple frameworks? Is that right? (I didnt migrate my projects to 2017)

mkosieradzki:
You can even have a single project with multiple frameworks and even architecutres and it works just great. In VS2017 you have much more manual control on .csproj file (it's way more simple and more convenient).

mbdavid:
Thats awsome, i hate this multiples solutions/projects doing same think. I will upgrade, thanks

Alexx999:
This PR removed assembly signing which makes resulting DLL incompatible with previous versions.

PR #661, opened 2017-07-22
fix #651

### LINKED ISSUE #651: FileDiskService in Shrink
in [LiteDB/Database/LiteDatabase.cs 234](https://github.com/mbdavid/LiteDB/blob/c071b627a7b6fed6986bf8e560c05f469b346435/LiteDB/Database/LiteDatabase.cs#L234)
FileDiskService created, but not used further

```c#
   // get temp disk based on temp file
   var tempDisk = new FileDiskService(tempFile, false);
    
   var reduced = _engine.Value.Shrink(password);
    
   // delete temp file
   File.Delete(tempFile);
```

schulz3000:
looks like all the temp stuff is not needed here.
UnitTests running without error over Shrink

### DIFF
--- a/LiteDB/Database/LiteDatabase.cs
+++ b/LiteDB/Database/LiteDatabase.cs
@@ -225,26 +225,7 @@ namespace LiteDB
         /// </summary>
         public long Shrink(string password)
         {
-            // if has connection string, use same path
-            if (_connectionString != null)
-            {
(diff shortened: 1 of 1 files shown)


mbdavid:
@schulz3000 thansk, but to fix #651 need pass tempDisk as parameter 8077c4e

PR #662, opened 2017-07-24
Enabling query contains to ignore case

jkvargas:
Just adding an option to ignore case when you are going Query.Contains =)


mbdavid:
Hi @jkvargas, thanks! I will merge soon.

PR #677, opened 2017-08-04
Adds optional parameter to dispose of a stream with database

Gusdor:
Adds optional parameter to dispose of a stream when the database is disposed. The extra parameter in the LiteDatabase constructor will communicate to consumers that the specified stream is not disposed by default.

I tried my best to respect the code conventions used in the solution but I couldn't find the contribution guidelines. Specifically, there is no unit test for this new parameter. Please let me know if this is an issue.


mbdavid:
Hi @Gusdor, thanks! I merge your PL in my dev branch because I will release 3.1.2 just fixing version number and lock service.

PR #698, opened 2017-08-17
ObjectId's Equals should return false when other is null.

vjacquet:
Current implementation of Equals for ObjectId throws NullReferenceException when other is null when it should return false (see https://msdn.microsoft.com/en-us/library/bsc2ak47(v=vs.110).aspx).
Added test to reproduce the issue then fixed the implementation.


mbdavid:
Thanks!

PR #705, opened 2017-08-22
Added Extension Methods for LiteStorage that resemble System.IO.File

Henry00IS:
I decided to keep them extensions so the programmer can identify that they are using the pre-existing methods made available by LiteStorage.

These extensions make it easier to read and write files as they remove the need for using streams just to get some string or byte array action.

`string ReadAllText(this LiteStorage self, string id)`
`string ReadAllText(this LiteStorage self, string id, Encoding encoding)`
`byte[] ReadAllBytes(this LiteStorage self, string id)`
`void WriteAllText(this LiteStorage self, string id, string filename, string contents)`
`void WriteAllText(this LiteStorage self, string id, string filename, string contents, Encoding encoding)`
`void WriteAllBytes(this LiteStorage self, string id, string filename, byte[] bytes)`
`void AppendAllText(this LiteStorage self, string id, string filename, string contents)`
`void AppendAllText(this LiteStorage self, string id, string filename, string contents, Encoding encoding)`

## Summary by CodeRabbit

- **New Features**
  - Added atomic text appending with optional encoding support.
  - Appends preserve existing content and file metadata, support large files, and handle concurrent writes safely.
  - Failed append and write operations roll back to prevent partial updates.
  - Added append-stream support with protection against use inside an active transaction.
  - Missing files report a clear `FileNotFoundException`.

- **Tests**
  - Expanded coverage for encodings, concurrency, metadata preservation, large files, transaction restrictions, and rollback scenarios.

### DIFF
--- a/LiteDB/Storage/LiteStorageExtensions.cs
+++ b/LiteDB/Storage/LiteStorageExtensions.cs
@@ -1,0 +1,164 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace LiteDB
+{
+    /// <summary><see cref="LiteStorage"/> extension methods that mimic <see cref="System.IO.File"/> methods.</summary>
+    public static class LiteStorageExtensions
+    {
+        /// <summary>Opens a text file, reads all lines of the file, and then closes the file.</summary>
+        /// <param name="self">The this pointer to the <see cref="LiteStorage"/> for the extension method.</param>
+        /// <param name="id">The identifier of the file to open for reading.</param>
+        /// <returns>A string containing all lines of the file.</returns>
+        /// <exception cref="FileNotFoundException">The file specified in id was not found.</exception>
+        public static string ReadAllText(this LiteStorage self, string id)
+        {
+            if (!self.Exists(id)) throw new FileNotFoundException("The file specified in id was not found.", id);
+            try
+            {
+                using (LiteFileStream stream = self.OpenRead(id))
+                using (StreamReader reader = new StreamReader(stream))
+                {
+                    return reader.ReadToEnd();
+                }
+            }
+            catch (Exception) { throw; }
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


JKamsker (dismissed):
Merge-readiness update for `12a15718d532`.

Changes needed before merge:

- Implement append without decoding and rewriting the existing file. The current ReadAllText-plus-WriteAllText approach changes prior bytes and can change the encoding; the review's default-append check rewrote existing UTF-16 content as UTF-8. Preserve existing bytes and relevant metadata, or split/defer append while landing the simpler helpers separately.
- Avoid loading the complete existing file into memory merely to append. Validate behavior for files spanning multiple storage chunks and for explicitly chosen encodings, including BOM handling.
- Add write-failure and large-file coverage, specifying which guarantees hold if an append fails. Preserve the passing basic text/binary and generic-ID cases and run current CI.

JKamsker:
All requested changes are addressed on `41660111`.

Summary:
- Synced the contributor branch with the latest `dev` (`a50661a9`) via merge.
- Added an atomic transactional `OpenAppend` path and switched both `AppendAllText` overloads to it.
- Append now preserves existing bytes and metadata, reuses only a partial tail chunk, avoids duplicate encoding preambles, and rolls back failed appends.
- Hardened stream failure handling so partial chunk-write failures are not retried during disposal, while ordinary invalid `OpenWrite` calls still finalize consistent metadata.
- Avoided append transaction lock inversion by rejecting caller-owned active transactions.
- Added coverage for both concurrent overloads, generic IDs, binary/text helpers, byte/metadata preservation, BOM behavior, multi-chunk files, encoder/write failures, overwrite failure consistency, and transaction rejection.

Validation:
- Focused net8.0 storage extension tests: 11 passed.
- Production library Release build: passed for netstandard2.0, net8.0, and net10.0.
- Changed-file size and diff checks: passed.
- Final GitHub CI: 45/45 jobs passed across the full OS/framework/architecture and repro matrix: https://github.com/litedb-org/LiteDB/actions/runs/34786196600

All review threads are resolved. The PR is mergeable with a clean merge state.

PR #726, opened 2017-09-09
Able to access property name in expression

Aimeast:
This PR will enable use `nameof` for write a expression.

eg. we can write an expression like this
```cs
using (var db = new LiteRepository(new MemoryStream()))
{
    db.Database
        .GetCollection<User>()
        .EnsureIndex(x => x.Name, true, x => $"LOWER($.{nameof(x.Name)})");
    db.Insert(new User { Name = "Tom", Email = "Tom@aa.xx" });
    var tom = db.First<User>(x => x.Name == "tom");
}
```

### DIFF
--- a/LiteDB/Database/Collections/Index.cs
+++ b/LiteDB/Database/Collections/Index.cs
@@ -26,11 +26,11 @@ namespace LiteDB
         /// <param name="property">Property linq expression</param>
         /// <param name="unique">Create a unique values index?</param>
         /// <param name="expression">Create a custom expression function to be indexed</param>
-        public bool EnsureIndex<K>(Expression<Func<T, K>> property, bool unique = false, string expression = null)
+        public bool EnsureIndex<K>(Expression<Func<T, K>> property, bool unique = false, Func<T, string> expression = null)
         {
             var field = _visitor.GetField(property);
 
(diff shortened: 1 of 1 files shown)


mbdavid:
Hi @Aimeast, but has a problem in this implementation: you are using only "nameof" to get property name and it's not always valid. If you User class is:

```C#
public class User
{
    [BsonField("user-name")]
}
```

In this case, your expression must be: 

`LOWER($.user-name)`

In this case, I think it's better create a "simple" linq visitor to convert:

```C#
db.Database
        .GetCollection<User>()
        .EnsureIndex(x => x.Name, true, x => x.Name.ToLower());

// into

db.Database
        .GetCollection<User>()
        .EnsureIndex(x => x.Name, true, "LOWER($.Name)");
```

I wrote days ago a similar linq visitor to convert C# Expression into Javascript string. The idea is the same:

https://github.com/mbdavid/DotVue/blob/master/DotVue/Utils/ExprJs.cs

Aimeast:
Yeah, I vote to linq converter.

JKamsker:
Thanks for the idea! v5 now translates the index lambda itself, so col.EnsureIndex(x => x.Name.ToLower()) creates a LOWER($.Name) index and respects [BsonField] names. That covers this without a string callback, so I'm closing it.

PR #728, opened 2017-09-11
Add Collection Swapping

Bustux:
I know this addition makes no sense for 99 % of all use cases.
But i would like to write extension Methods for the LiteCollection in my own project and in this case it is impossible to easy swap between parametric types. So i decided to add this feature and hope it will be accepted.


JKamsker:
Thanks for the contribution! The same thing is available through db.GetCollection<OtherType>(col.Name), or db.GetCollection(col.Name) for BsonDocument, because a collection is only a name over the same engine, so I'm closing this.

PR #758, opened 2017-10-09
Make upsert return true if updating.

drjonnicholson:
We're looking at switching from SQLite to LiteDB in one of our products, but we found that `false` is returned when doing an `upsert` if the document was updated rather than inserted.

We would argue that the user only cares if an `upsert` worked, and doesn't care if it was actually inserted or updated, so `true` should be returned in both insert and update cases.

This pull request changes the logic of `upsert` in the engine so that the return value of up `UpdateDocument` is accounted for in the `count` variable.

### DIFF
--- a/LiteDB/Engine/Engine/Upsert.cs
+++ b/LiteDB/Engine/Engine/Upsert.cs
@@ -33,12 +33,24 @@ namespace LiteDB
 
                 foreach (var doc in docs)
                 {
-                    // first try update document (if exists _id)
-                    // if not found, insert
-                    if (doc["_id"] == BsonValue.Null || this.UpdateDocument(col, doc) == false)
+                    if (doc["_id"] == BsonValue.Null)
                     {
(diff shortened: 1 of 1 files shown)


mbdavid:
Hi @drjonnicholson, `Upsert` always "work". Upsert will Insert or Update a document. If not work (by an infrastructure problem), will throw an exception.
So, I use use bool (or int) to know if was insert or update.
In you code, if you test `db.Upsert(...) == true` it's always true. So, use as a `void` method (you can use inside a try/catch to get any error)

drjonnicholson:
I see your point of view. 

In which case I think adding a comment like "Returns true if insert entity or false if update entity" to the `Collections@Upsert` summary, as taken from the `LiteRepository@Upsert` summary, would help people avoid the same misconception I had, i.e. thinking `upsert` had failed because it returned false.

drjonnicholson:
Closing - no need for this PR and it's quite old now!

PR #798, opened 2017-11-12
create predicate in LiteQueryable.Exists

Ryumkin:
Hi.
I think so method "Exists" need implementation in predicate.


mbdavid:
Hi @Ryumkin, I will add this parameter too.. but to all "fetch" methods

PR #813, opened 2017-11-30
Fix constructor ignoring log level if Logger passed. Fixes #776

### LINKED ISSUE #776: Logging not working at all?
I've just tried to introduce LitDB logging of my Xamarin Forms app, based on this https://github.com/mbdavid/LiteDB/wiki/Connection-String
I.e.:

`new LiteRepository("filename=/data/user/0/com.mycompany.myapp/files/LocalDb/DB_1.0_Logged_User.db;log=255;journal=false")`

255 is based on `Logger.FULL`

However I didn't see completelly no logs, during selecting/insering/updating objects in LiteDB.

Any idea?

parabola949:


mbdavid:
Hi @parabola949, thanks for that!

PR #852, opened 2017-12-23
reduce exception throw when use complex linq query

JasonGrass:
Hi ,  
Actually，it,s not a good idea to use exception for flow control. 
This update can reduce exception throw when use complex linq query.   
When I open  ' break all exception when Common Language Runtime Exceptions happen' in visual studio , it always break when I use complex linq .


mbdavid:
Thanks @JasonGrass

PR #854, opened 2017-12-28
Additional constructor in LiteRepository and EditorConfig file

axelgenus:
I added an EditorConfig file in order to set the indentation globally (works on VS2017+ and with ReSharper).

Also I added a LiteRepository constructor which accepts a pre-existent LiteDatabase instance instead of creating a new one (useful in case of an encrypted database or of needing logging for debugging purposes). The new constructor also gives the option to dispose the database's instance together with the repository's. Old constructors have been redirected to the new one for consistency.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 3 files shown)


axelgenus:
It seems like the CI is broken: the tests are in the path
"C:\projects\litedb\LiteDB.Tests\bin\Debug\netcoreapp2.0\LiteDB.Tests.dll" but it looks for them in
"C:\projects\litedb\LiteDB.Tests\bin\Any CPU\Debug\netcoreapp2.0\LiteDB.Tests.dll".

mbdavid:
Hi @axelgenus, thanks. CI are with some problems... I will fix this later

PR #883, opened 2018-01-25
Fix "new ConnectionString()" not containing default values

PathogenDavid:
For your consideration, this pull request fixes a bug I found when evaluating LiteDB for my project.

If you create a `ConnectionString` using the parameterless constructor (with the intent to fill out the properties manually), then the resulting `ConnectionString` will not contain the appropriate defaults. For example:

```c#
var connectionString = new ConnectionString()
{
    Filename = "test.db",
    Mode = FileMode.Exclusive
};

// connectionString.LimitSize == 0, but we should expect it to be long.MaxValue!
```

This pull request fixes this change by using the parameterless constructor to populate the default values. Additionally, I reworked `ConnectionString(string)` to use the parameterless constructor for the defaults to avoid error in differing defaults in the future.

I've also updated `ConnectionString_Tests` to test the default values (for both constructors.) I also added some missing tests for `Async` and `Upgrade`, and a meta test to prevent test rot.

Additionally, I introduced `LiteDbBuildConfiguration`, which provides insight into how the LiteDB assembly you're using was built. I originally introduced this before I realized the tests are only run against the .NET Standard 2.0 version of LIteDB (so I thought I needed to support the `Async` property only being present in .NET standard builds.) So if you don't plan on ever running the tests against other build configurations, it could be removed.

Finally, I updated some of the documentation on ConnectionString to fix missing/incorrect values and consistency. I can/will update the wiki as well when these changes are merged.

Thanks! Let me know if you have any questions or concerns about my changes.


mbdavid:
Hi @PathogenDavid, thanks for your PR. I didn't test this string connection. I will accept but, as you mention, `LiteDbBuildConfiguration` is not needed for tests. Also, to simplify, properties can be setted direct on property declaration (C# 6) like (I lready updated in v5 branch here: https://github.com/mbdavid/LiteDB/blob/v5/LiteDB/Utils/ConnectionString.cs):

```
public TimeSpan Timeout { get; set; } = TimeSpan.FromMinutes(1);
```

Did you want update this or I change later?

PathogenDavid:
Oh duh, I don't know why I didn't use property initializers in the first place. I've resubmitted this PR as #884 with that change, as well as with the `LiteDbBuildConfiguration` stuff removed.

I was originally going to look into making the tests run against all target frameworks as a separate PR (since it seems like they really ought to), but it seems Visual Studio doesn't play nice with doing that. So I've removed `LiteDbBuildConfiguration` and the logic associated with it.

PR #884, opened 2018-01-27
Fix "new ConnectionString()" not containing default values

PathogenDavid:
This an amended version of #883 with suggested property initializer usage and unnecessary LiteDbBuildConfiguration removed.


mbdavid:
Thanks @PathogenDavid!!

PR #890, opened 2018-02-04
Microfix for ToString method within QueryIn class.

trueromanus:
The **ToString** method in the **QueryIn** class returns only a collection of types with no values. The remaining **Query** classes return all required values as a string. I use this method for the test, and for QueryIn it's not impossible.
I added output of all values, separated by a comma. Maybe it will be interesting to you.
Sorry for poor english.


mbdavid:
Thanks @trueromanus!

PR #903, opened 2018-02-11
Add SingleOrDefaultById to LiteRepository

mstum:
`LiteRepository.SingleById` is super useful, but I might have cases where the entity either exists or doesn't, in which case `System.Linq.Enumerable.Single()` will throw an exception.

This adds `SingleOrDefaultById` which uses `System.Linq.Enumerable.SingleOrDefault()` instead.


JKamsker:
Superseded by #2878, which imports this contribution’s history, ports optional ID lookup to the current query/repository APIs, and adds regression coverage. #2878 has now been merged into dev. Because it preserves this PR’s commit history, GitHub automatically marked this original PR as merged too; no separate merge was performed for #903.

PR #904, opened 2018-02-11
Query.StringEQ / QueryStringEquals with StringComparison

mstum:
I have several scenarios where I need to fetch an object not by its ID, but by another field. The problem with `Query.EQ` is that it does a straight comparison, which means that strings are compared case-sensitive. Very often, the incoming value might be differently cased than the one stored, which means that I'd have to make sure to always keep them the same (e.g., calling `ToUpperInvariant` when saving and querying.

This adds a `QueryStringEquals` Query and `Query.StringEQ` method which is specifically for strings and accepts a `StringComparison`, which allows passing in e.g., `StringComparison.OrdinalIgnoreCase` for a case insensitive string comparison.

### DIFF
--- a/LiteDB.Tests/Repository/Repository_Tests.cs
+++ b/LiteDB.Tests/Repository/Repository_Tests.cs
@@ -82,7 +82,8 @@ namespace LiteDB.Tests.Repository
                 // query with StringEQ
                 Assert.IsNotNull(db.Query<RCustomer>().Where(Query.EQ(nameof(RCustomer.CustomerName), "John")).FirstOrDefault());
                 Assert.IsNull(db.Query<RCustomer>().Where(Query.EQ(nameof(RCustomer.CustomerName), "john")).FirstOrDefault());
-                Assert.IsNotNull(db.Query<RCustomer>().Where(Query.StringEQ(nameof(RCustomer.CustomerName), "john", ignoreCase: true)).FirstOrDefault());
+                Assert.IsNotNull(db.Query<RCustomer>().Where(Query.StringEQ(nameof(RCustomer.CustomerName), "john", StringComparison.OrdinalIgnoreCase)).FirstOrDefault());
+                Assert.IsNull(db.Query<RCustomer>().Where(Query.StringEQ(nameof(RCustomer.CustomerName), "john", StringComparison.Ordinal)).FirstOrDefault());
 
                 // check is exists
                 var r2 = db.Query<ROrder>()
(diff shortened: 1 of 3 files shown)


mbdavid:
Hi @mstum, thanks. You are not using index as best way, because you are using full scan index search. To this search works as needed, database will need support collation, when all database, collection or specific index are created with this option. It's on my plans for next version.
With current version, if you need ignore case, I recommend you create a lower case expression and use lowecase string, like this:

```
col.EnsureIndex("Name", "LOWER($.Name)");
```

mstum:
Thanks, that'll work as well. I'd do uppercase though, [lowercase may not roundtrip properly](https://docs.microsoft.com/is-is/visualstudio/code-quality/ca1308-normalize-strings-to-uppercase) in all locales.

alfcas:
I changed QueryVisitor this way:
line 138:
                    // StartsWith
                    if (method == "StartsWith")
                    {
                        var value = this.VisitValue(met.Arguments[0], null);
                        //return Query.StartsWith(this.GetField(met.Object, prefix), value);
                        if (met.Arguments.Count == 1)
                        {
                            return Query.StartsWith(this.GetField(met.Object, prefix), value);
                        }
                        else if (met.Arguments.Count == 2)
                        {
                            return Query.StartsWith(this.GetField(met.Object, prefix), value,
                                GetParamValue<StringComparison>(met.Arguments[1]));
                        }
                        else if (met.Arguments.Count == 3)
                        {
                            return Query.StartsWith(this.GetField(met.Object, prefix), value,
                                GetParamValue<bool>(met.Arguments[1]),
                                GetParamValue<System.Globalization.CultureInfo>(met.Arguments[2]));
                        }
                        else
                            throw new Exception("Parameter list too long");
                    }

...
line 372

        private P GetParamValue<P>(Expression expression)
        {
            var objectMember = Expression.Convert(expression, typeof(P));
            var getterLambda = Expression.Lambda<Func<P>>(objectMember);
            var getter = getterLambda.Compile();
            return getter();
        }
        private object GetParamValue(Expression expression)
        {
            var objectMember = Expression.Convert(expression, typeof(object));
            var getterLambda = Expression.Lambda<Func<object>>(objectMember);
            var getter = getterLambda.Compile();
            return getter();
        }

...
and implementint on Query.cs 3 different methods 
...
and changing QueryStartsWith.cs to distinguish the 3 different calls

JKamsker:
Thanks for this, and for the uppercase tip! v5 added database collations (case-insensitive by default, configurable with Collation= in the connection string), and UPPER()/LOWER() work in queries and index expressions, so this is covered and I'm closing it.

PR #911, opened 2018-02-15
Created IncludeAll method

nelsoncmartinsf:
Created an "IncludeAll" method to include all references to other tables.

### DIFF
--- a/LiteDB.Tests/Database/DbRef_Include_Tests.cs
+++ b/LiteDB.Tests/Database/DbRef_Include_Tests.cs
@@ -173,5 +173,90 @@ namespace LiteDB.Tests.Database
 
             }
         }
+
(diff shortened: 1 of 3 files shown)


mbdavid:
Hi, I don't get it why change `GetField` method and do not use `FieldName` from member mapper.

Also, will be great an unit test included.

I'm do not write any code-style rule or setting editor config yet, but I recommend to follow code style rules, like:
- always use `var`
- any public method must have, at least, `<summary>` (sometimes in private method I add too)
- inside code, comments must be single line and always use lowercase (Capitalize only in `<summary>`)
- lambda expression with single char variable like:  `items.Query(x => x.Name)` (i always use `x` first)
- no extra spaces in `(`, `)`

Issue #908, opened 2018-02-14
PlatformNotSupportedException in LiteEngine ctor FileDiskService.Lock on Mac OS X using .Net Core 2.1.4


ryanseghers:
I'm trying to use LiteDB on Mac OS X 10.12.6 (Sierra) using .Net Core 2.1.4. I am getting an exception:

Unhandled Exception: System.PlatformNotSupportedException: Locking/unlocking file regions is not supported on this platform. Use FileShare on the entire file instead.
							at System.IO.FileStream.LockInternal(Int64 position, Int64 length)
							at LiteDB.FileHelper.TryExec(Action action, TimeSpan timeout)
							at LiteDB.StreamExtensions.TryLock(FileStream stream, Int64 position, Int64 length, TimeSpan timeout)
							at LiteDB.FileDiskService.Lock(LockState state, TimeSpan timeout)
							at LiteDB.LiteEngine..ctor(IDiskService disk, String password, Nullable`1 timeout, Int32 cacheSize, Logger log, Boolean utcDate)
							at LiteDB.LiteDatabase.<>c__DisplayClass11_0.<.ctor>b__0()
							at LiteDB.LazyLoad`1.get_Value()
							at LiteDB.LiteCollection`1.EnsureIndex(String field, String expression, Boolean unique)

See the Dec 12 2016 post in this discussion: https://github.com/dotnet/corefx/issues/5964
If I understand this discussion correctly then they have specifically decided not to support this particular locking behavior on Mac OS X, so LiteDB as is will never work.

I would like to use LiteDB on Mac OS X. Am I understanding the issue correctly? Is there any solution to this?

mbdavid:
Hi @ryanseghers, OS X do not  support file lock, so you cann't use LiteDB with lock support (no multi process access). But you can still use LiteDB in multi thread model, using: "Mode=Exclusive" in connection string (exclusive mode do not lock file because open file in exclusive mode

mstum:
FWIW, there is no way to just see if File Locks are supported in .net Core (e.g., no static FileStream.CanLock), the only option apart from hardcoding a list of OS platforms that support it would be to try to open the file during the constructor/init, catch the PlatformNotSupportedException and change the FileMode to .Exclusive.

Issue #787, opened 2017-10-31
Unable to write to a DB file on a macOS


DeeGeeMan:
I've tried using both LiteRepository and LiteDatabase to create and open a new DB, but I am getting an error when calling Insert.

**Error:**
Exception has occurred: CLR/System.PlatformNotSupportedException
An unhandled exception of type 'System.PlatformNotSupportedException' occurred in LiteDB.dll: 'Locking/unlocking file regions is not supported on this platform. Use FileShare on the entire file instead.'
   at System.IO.FileStream.LockInternal(Int64 position, Int64 length)
   at LiteDB.FileHelper.TryExec(Action action, TimeSpan timeout)
   at LiteDB.StreamExtensions.TryLock(FileStream stream, Int64 position, Int64 length, TimeSpan timeout)
   at LiteDB.FileDiskService.Lock(LockState state, TimeSpan timeout)
   at LiteDB.LiteEngine..ctor(IDiskService disk, String password, Nullable`1 timeout, Int32 cacheSize, Logger log)
   at LiteDB.LiteDatabase.<>c__DisplayClass11_0.<.ctor>b__0()
   at LiteDB.LazyLoad`1.get_Value()
   at LiteDB.LiteCollection`1.Insert(T document)
   at LiteDbDemo.Program.Main(String[] args) in /Users/grantd/Projects/Ecs/LiteDbDemo/Program.cs:line 20

mbdavid:
Hi @Devin-Grant, I didn´t know about this macOS limitation (lock/unlock regions). In this case, try use `mode=Exclusive` in your [connection string](https://github.com/mbdavid/LiteDB/wiki/Connection-String).

Devin-Grant:
Thant works!

avfranco-br:
Hi guys, I had the same issue above, used the Mode=Exclusive, however when trying to insert a new document (Employees.Insert(Employee)) getting the error - System.ArgumentNullException: Value cannot be null.
Parameter name: filename
   at LiteDB.FileDiskService..ctor(String filename, FileOptions options)
   at LiteDB.LiteDatabase.<>c__DisplayClass11_0.<.ctor>b__0()
   at LiteDB.LazyLoad`1.get_Value()
   at LiteDB.LiteCollection`1.Insert(T document)
   at LiteDBSample.Program.Main(String[] args) in

Any thoughts? Many thanks, Alexandre

lazlo-bonin:
I get the same issue as above.

[...]

PR #912, opened 2018-02-16
If FileStream.Lock is not supported, point towards the solution

mstum:
Ran into this issue on OS X, same as #787 and #908 

Unfortunately, .net Core has no way to see if FileStream.Lock is supported, the code [just throws an Exception](https://github.com/dotnet/corefx/blob/cbdc1817946b44b747c481a3782450d8d2a54942/src/Common/src/CoreLib/System/IO/FileStream.OSX.cs).

One option would be to try to open and Lock the file during `FileDiskService.Initialize`, and change to Exclusive if it throws, but that would be a bigger functional change than just catching and showing the message.

### DIFF
--- a/LiteDB/Utils/Extensions/StreamExtensions.cs
+++ b/LiteDB/Utils/Extensions/StreamExtensions.cs
@@ -48,6 +48,10 @@ namespace LiteDB
             {
                 return false;
             }
+            catch (PlatformNotSupportedException pex)
+            {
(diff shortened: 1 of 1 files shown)


mbdavid:
Hi @mstum, thanks. I agree with you, there is better show a user message to always check this in initialize.

In v5 I will use, when there is no `FileStream.Lock` and external file lock.

PR #913, opened 2018-02-18
Update Update.cs

ghost:
Gets Update Simpler


JKamsker:
Thanks for the contribution! The v4 branch only gets critical fixes now, and v5+ has UpdateMany(x => new T { ... }, predicate), which updates all matching documents in the engine in one transaction, so I'm closing this.

PR #916, opened 2018-02-18
Added notifications to update and insert

SeanMollet:
I needed the DB to throw an event on Update and Insert, so I've implemented a fairly simple notification system. You're welcome to add it back if you like it.


JKamsker:
Thanks for this, and sorry it waited so long! v5 rewrote the engine's write and transaction paths, so this can't be applied anymore. Change notifications are tracked in #2786, where the plan is to raise them only after a commit, so rolled-back writes aren't reported; ideas and feedback are welcome there.

PR #928, opened 2018-03-06
Fixed not generic type IDictionary BsonValue serialize

Jokder:
If obj type is IDictionary but not generic type,this will cause the type.GetTypeInfo().GetGenericArguments() is empty and throw a System.IndexOutOfRangeException: Index was outside the bounds of the array.So add a generic type check.

### DIFF
--- a/LiteDB/Mapper/BsonMapper.Serialize.cs
+++ b/LiteDB/Mapper/BsonMapper.Serialize.cs
@@ -102,7 +102,8 @@ namespace LiteDB
                     type = obj.GetType();
                 }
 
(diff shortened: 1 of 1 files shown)


Jokder:
finally i found type has no IsGenericType property,so i changed to GetTypeInfo().IsGenericType to made git verify passed.

JKamsker:
Thanks for this fix! Current dev (v6) already handles non-generic dictionaries the same way: the mapper only reads generic arguments when the type is generic and falls back to object otherwise (BsonMapper.Serialize.cs). The v4 file this PR changes no longer exists, so I'm closing it.

PR #933, opened 2018-03-11
Fix indentation

kveretennicov:


JensSchadron:
It's a pretty late answer, but thanks for fixing this.

PR #944, opened 2018-03-24
Enhanced connection string parsing

tdinucci:
Enhanced connection string parsing.  
Meaningful error messages at parse time and allow for passwords to be formed from string literals which include previously unallowed characters, such as semi-colons.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection-string parsing for quoted values, escaped characters, whitespace, Windows paths, and custom options.
  * Malformed connection strings now produce clearer format errors, while repeated options consistently use the final value.
  * Settings-only connection strings remain supported.
  * Connection-string options now preserve quoted content, including semicolons, quotes, and spaces.

* **Documentation**
  * Updated the `open` command help example to use the correct `initial size` connection-string parameter.

### DIFF
--- a/LiteDB.Tests/Database/ConnectionString_Tests.cs
+++ b/LiteDB.Tests/Database/ConnectionString_Tests.cs
@@ -17,6 +17,104 @@ namespace LiteDB.Tests.Database
         }
 
         [TestMethod]
+        public void ConnectionString_Passwords()
+        {
+            var parsed = new ConnectionString(@"filename=demo;password=""ab\""|    c;785;""");
+            Assert.AreEqual("demo", parsed.Filename);
+            Assert.AreEqual(@"ab""|    c;785;", parsed.Password);
+
+            var nonParsed = new ConnectionString
+            {
+                Filename = "demo",
+                Password = @"ab""|    c;785;"
+            };
+
+            Assert.AreEqual(parsed.Filename, nonParsed.Filename);
+            Assert.AreEqual(parsed.Password, nonParsed.Password);
+
+            parsed = new ConnectionString(@"filename=demo;password=ab\""123");
+            Assert.AreEqual(@"ab""123", parsed.Password);
+
+
+            parsed = new ConnectionString(@"filename=demo;password=""ab;123"";");
+            Assert.AreEqual(@"ab;123", parsed.Password);
+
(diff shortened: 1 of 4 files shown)


JKamsker (changes requested):
Merge-readiness update for `d87944ed75f5`.

Changes needed before merge:

- Preserve custom options exposed through the public ConnectionString indexer, or obtain explicit approval for a documented breaking change. The reviewed dev baseline accepts a custom key and exposes its value through the indexer; this PR's fixed whitelist throws instead.
- Separate syntax validation from recognition of built-in options so adding a supported option does not require another parser compatibility break. Preserve current settings-only strings and built-in memory settings.
- Consolidate the compatibility tests with #2757 and #2745: custom options, malformed tokens, equals-sign paths, quoted/escaped values, Windows/UNC paths, whitespace, duplicate keys and round-trip settings.

The full 601-test .NET 8 run passed during conflict resolution, but did not cover the custom-option regression. The `autorebuild` spelling without a hyphen is not an existing built-in alias on dev; it should not be cited as the baseline regression.

JKamsker (approved):
The requested compatibility concerns are addressed on 6b853100: custom options are preserved, syntax validation is independent of the built-in option set, malformed empty segments remain rejected, and the parser tests cover custom and duplicate options alongside the merged #2757 compatibility suite. The shell example now uses the supported initial size key.

JKamsker:
Updated the PR head and addressed all review feedback:

- merged the latest `dev` (`a50661a9`) into the contributor branch
- removed the parser allow-list so custom indexer options remain compatible while retaining strict malformed-input validation
- added coverage for custom options, duplicate-key last-value behavior, and empty option segments; the merged #2757 suite continues to cover equals-sign paths and compatibility cases
- corrected the shell example to use `initial size`

Validation:
- focused .NET 8 connection-string/#2211 tests: 66 passed
- Release solution build with `TestingEnabled=true`: succeeded for all configured targets
- full .NET 8: 875 passed, 7 skipped
- full .NET 10: 875 passed, 7 skipped
- repro-runner: 18 passed
- changed-file size check and `git diff --check`: passed

The net462/net481 test hosts could not execute locally because this Linux environment does not have Mono; both target assemblies compiled successfully. CodeRabbit completed on `6b853100` with no actionable comments.

JKamsker:
CI follow-up: approved GitHub Actions run `34784029523` completed successfully with all 45 jobs passing.

Issue #768, opened 2017-10-15
Null Reference on Insert


ReiSchneider:
I  ran into this problem where in I get a null reference exception on insert.

I have two object Models UserInfo and UserConfig. On the first trial, UserConfig references a UserInfo instance

```
public class UserConfigObject : IUserConfig
    {
        BsonRef("userInfo")]
        public IUserInfo UserInfo { get; set; }

        public string AssignedJob { get; set; }

        public string[] QueueItems { get; set; }

    }
```

```
public class UserInfoObject : IUserInfo
    {
        [BsonId]
        public int ID { get; set; }

        public string Name { get; set; }
        public string Username { get; set; }

        public string IPAddress { get; set; }
    }
```

And a method to insert the data into the database

```
public void AddUser(IUserConfig user)
        {
            var uconCollection = DatabaseInstance.GetCollection<IUserConfig>("userConfig");
            var uinCollection = DatabaseInstance.GetCollection<IUserInfo>("userInfo");

            uinCollection.Insert(user.UserInfo);
            uconCollection.Insert(user);

        }
```

This set up works fine but when I try to change the reference to UserInfo references UserConfig

```
public class UserInfoObject : IUserInfo
    {
        [BsonId]
        public int ID { get; set; }

        public string Name { get; set; }
        public string Username { get; set; }

        public string IPAddress { get; set; }

        [BsonRef("userConfig")]
        public IUserConfig UserConfig { get; set; }
    }
```

```
public class UserConfigObject : IUserConfig
    {
        [BsonRef("userInfo")]
        public IUserInfo UserInfo { get; set; }

        [BsonId(true)]
        public int ConfigID { get; set; }

        public string AssignedJob { get; set; }

        public string[] QueueItems { get; set; }

    }
```

With a method call for

```
public void AddUser(IUserInfo user)
        {
            var uconCollection = DatabaseInstance.GetCollection<IUserConfig>("userConfig");
            var uinCollection = DatabaseInstance.GetCollection<IUserInfo>("userInfo");

            uconCollection.Insert(user.UserConfig);
            uinCollection.Insert(user);
        }
```

It no longer works, it throws an System.NullReferenceException: 'Object reference not set to an instance of an object.' on uinCollection.Insert(user);

Either v3 or v4, it doesn't work with the latter set up

mbdavid:
[...]

PR #951, opened 2018-04-09
Proposed fix for Issue #768

BHelfmann:
As I got the same error as the user in Issue #768, but with a DbRef pointing to an interface class, I made a fix that works for me. It might be beneficial as a starting point for a real fix.

The problem seems to be that the DbRef deserialization has to guess the type of the entity to load from its declared base type. If that one is an interface or abstract class, we're in trouble, because Activator.CreateInstance doesn't work on those.

There are basically two ways that I imagine will fix the problem:

1. Save the type of the DbRef target object in the reference itself, so we have a precise expression of Type and don't need to guess.
Pro: access patterns and therefore performance don't change (much).
Con: we specify the type twice, in the DbRef and in the actual entity.

2. Get the type from the DbRef target entity itself.
Pro: we specify the type only once, in the entity.
Con: there's a performance penalty getting the type from the entity, as well as an architectural breach, reaching out from BsonMapper to the database.

The proposed solution is option (1). Use it at your discretion.

Thanks for the work you're doing with LiteDB! It's great!

### DIFF
--- a/LiteDB/LiteDB.nuspec
+++ b/LiteDB/LiteDB.nuspec
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata>
+    <id>LiteDB</id>
+    <version>4.1.2-patched</version>
+    <title>LiteDB</title>
+    <authors>Maurício David</authors>
+    <owners>Maurício David</owners>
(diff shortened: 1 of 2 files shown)


JKamsker:
Thanks for digging into this and proposing a fix! LiteDB v5 and current dev now store the concrete type as $type on DbRefs whenever it differs from the declared type, for single references and lists, which covers interface and abstract references. The PR also targets v4 code and adds a patched nuspec, so I'm closing it.

Issue #1058, opened 2018-08-13
Upsert array and Insert array memory leaks


falahati:
Hey David, me again :)

Take a look at the following results:

> Insert Test:
> -- Memory Used: 36MB - Elapsed: 81163ms
> Insert Array Test:
> -- Memory Used: 116MB - Elapsed: 1981ms
> Insert Bulk Test:
> -- Memory Used: 35MB - Elapsed: 1987ms
> Insert or Update Test:
> -- Memory Used: 36MB - Elapsed: 83342ms
> Upsert Test:
> -- Memory Used: 116MB - Elapsed: 2179ms
> Manual Upsert Test:
> -- Memory Used: 35MB - Elapsed: 145922ms
> Finished.

A sample project is attached:
[LiteDBUpsertLeak.zip](https://github.com/mbdavid/LiteDB/files/2281749/LiteDBUpsertLeak.zip)

Now, don't underestimate the 80MB difference, that's for 100,000 rows. Increasing the number to 1M rows and you get this:

> Insert Array Test:
> -- Memory Used: 1172MB - Elapsed: 21948ms
> Insert Bulk Test:
> -- Memory Used: 24MB - Elapsed: 22438ms
> Upsert Test:
> -- Memory Used: 1172MB - Elapsed: 24621ms
> Finished.

falahati:
The proposed new method should kinda solve the problem with Upsert. I didn't actually go through all parts of the code as I was expecting to get lost, but simply batching the array seems to solve the problem. I have no idea why tho, so the underlying problem is still there.

Do you maybe cache something in transactions? It seems like a common problem with long transactions.

lbnascimento:
Hi! With the objective of organizing our issues, we are closing old unsolved issues. Please check the latest version of LiteDB and open a new issue if your problem/question/suggestion still applies. Thanks!

PR #1059, opened 2018-08-13
A new UpsertBulk method added to solve the problem with high memory usage of Upsert

falahati:
#1058


falahati:
Sorry about the `Using System;` part, my VS or Git client probably don't like the encoding of files (BOM?) or something like that and I completely missed that in commit review.

mbdavid:
Hi @falahati, after a long while, I'm backing to code... thanks for your contribution. I accept your code.

PR #1109, opened 2018-10-22
Introduce ITypeNameBinder and default implementation

### LINKED ISSUE #1108: Customize how _type is handled
Currently the `_type` property always contains the assembly qualified name of the type. This means that if something changes between versions (e.g. an assembly or type is renamed), existing document can no longer be deserialized.

Is there a way to use custom type/name mappings instead? From what I see in the code, I don't think so, but it would be a useful addition.

JSON.NET offers an extension point for this: the `ISerializationBinder` interface, which lets you map a type to a name and vice versa. Is this something you'd consider adding to LiteDB? It would be very useful to make it easier to evolve apps with existing data.

thomaslevesque:
Fixes #1108


mbdavid:
Hi @thomaslevesque, interesting idea. I never faced with this problem. You code sound nice. Also, I think will be nice adding some `BsonBinderNameAttribute` class to decorate class, like:

```
[BsonBinderName('User')]
public class NewUserClass
{
    ....
}
```

thomaslevesque:
Hi @mbdavid,

Yes, I guess the attribute could be useful too. Should probably be in a separate PR, though.

thomaslevesque:
Hi @mbdavid,

Is there anything blocking this PR from being merged?
I can add support for attributes if you like, I just thought it would make sense to do it in another PR.

PR #1137, opened 2018-12-04
Support FileStorage ids with spaces 

MattyBoy4444:
This allows better support for using the the path as the id. Otherwise, if an id has a space an InvalidFormat exception will be thrown.


mbdavid:
Thanks @MattyBoy4444

PR #1140, opened 2018-12-08
Fixes and trivial performance improvements for v4

hex11:
I wasted a whole day...

### DIFF
--- a/LiteDB.Tests/Engine/Shrink_Tests.cs
+++ b/LiteDB.Tests/Engine/Shrink_Tests.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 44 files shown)


mbdavid:
Hi @hex11, thanks a lot of your code. I was not a waste time, it's a great community helps :)

I have a question about your code:

- Why convert BsonReader/Writer into `struct` as better than a class? I always make me this same question when I create any data structure..... all docs about this tells me that I must use struct only in some cases, like immutable data and small size (to avoid copy data structure as parameters)

hex11:
I'm happy to see your reply :)

I noticed new (Bson/Byte)(Reader/Writer) instances will be created in some kinds of (de)serializing,
so I converted `class` to `struct` to make them be allocated on the thread stack instead of the GC heap, and pass them as argument by `ref` (it passes the pointer of the `struct` on stack, just like passing a `class` object, but the pointer points to stack. So there's no struct copying. Actually, it won't work without `ref`, because they have some states to be changed (e.g. `ByteReader._pos`)).

But I'm not sure if I should do that, because there are so many appearances of `ref` in the code and the performance improvement is minor.

hex11:
Close it for now because v5 is on the way and there are maybe too many hacks in this PR.

But is there any plan to release new v4 for bug fixes? If there is, you can pick the first 3 commits here. They are the actual fixes.

By the way, new features in v5 are awesome. Please keep developing it. 👍

PR #1191, opened 2019-03-08
Improve performance by 15-20%

Romasz:
As you recognize the type when BsonValue is created, you can hold it in dedicated variable - having a set of such variables would allow to get rid of boxing/unboxing operations on RawValue and optimize the code. The only drawback is that it will cost few bytes more, though on the other hand some methods/classes can be made simpler. 

This can also be done by holding only bytes and using pointer casing, in this case the class won't be heavier but will need unsafe code (not sure if you want that one).

Summary of changes:
 - got rid of boxing/unboxing of RawValue
 - BsonArray doesn't need index as a key - this also makes database little lighter, (needed to implement new type for backward compatibility)
 - reorganized BsonDocument & BsonArray classes
 - reorganized how Length is calculated
 - some more little changes

As I've tried the the Demo I've got such timings:
  - on SSD -> old version 11416 ms, new version 9181 ms.
  - on HDD -> old version 15460 ms, new version 12005 ms.


mbdavid:
Hi @Romasz, thanks for your contribution. I was checking your code and was based on avoid cast in `BsonValue.RawValue`. But, use a specific variable for each data type will grow lot's of memory usage for this. This `BsonValue` is the most used memory class in any test I did.

I checked how JSON.NET implement this and they use 1 class per data type... I think this is the best solution for this case: will avoid cast and still use same memory space.

Romasz:
@mbdavid I thought about this also a little and this may be also done two different ways, either by using unsafe code and pointer to data or with the new Span class there is option to cast data to different types, so one byte array can be used to hold different types depending on needs.

Nevertheless getting rid of unnecessary boxing/unboxing gives nice performance boost. I concur that separate classes may be the best solution here.

marius-klimantavicius:
Alternatively to using unsafe/pointers and span, you could use "semi-"unsafe struct with explicit layout and FieldOffset(0). Microsoft uses this trick in sql client:
https://referencesource.microsoft.com/#System.Data/fx/src/data/System/Data/SqlClient/SqlBuffer.cs,21b924f96987213c,references

JKamsker:
Thanks a lot for the benchmarks and the effort here! This is based on the 2019 v5 alpha, and the engine, file readers and BSON classes have changed too much since then for it to be rebased. The maintainer at the time also preferred a different design for BsonValue (per-type classes rather than per-type fields). Performance work is now tracked in #2832, so I'm closing this PR.

PR #1200, opened 2019-04-02
Improve Dispose pattern implementation

OlegLevitsky:
Fixes https://github.com/mbdavid/LiteDB/issues/742

According to MS [documentation](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/b1yfkh5e(v%3dvs.71)) any managed calls can't be called from the finalizer


mbdavid:
Thanks @OlegLevitsky

IvanKonov:
When will it appear in NuGet?

PR #1218, opened 2019-05-09
Fix mixed leading whitespace in sample code

patridge:
Just noticed that there were mixed tabs and spaces in the README sample code, causing it to misalign. Several of the blank lines also started with tabs, so I cleaned them up as well.

Screenshot of README with misaligned `var customer` line


JensSchadron:
@mbdavid Could you merge this as it only affects the README?

JensSchadron:
Thanks for cleaning this up.

PR #1226, opened 2019-05-20
Allows string identification fields

falahati:
This change allows for a `string` property to act as the `_id` field similar to an `Int64` property.

This feature allows for more flexibility for identification fields especially in cases in which the Id field is not necessarily always an integer.

The only downside to this might be the fact that one can insert a bad id manually and therefore pushes the current column sequence way further. However, this is also the case for all integer identification fields as well.


JKamsker:
Thanks for the PR! v5 reworked auto-id (BsonAutoId, with the sequence handled in the engine), so this v4 change no longer applies. Configurable auto-ids, including string ids, are tracked in #2837, where the current direction is a pluggable id generator. Closing this one in favour of that.

Issue #1224, opened 2019-05-19
Implicit BsonValue from ulong operator casts to double instead of long.


Quahu:
Hello, is there any particular reason why the following implicit conversion operator casts the `UInt64` to `double` instead of `long`? 
I've ran into this while attempting to fetch a document via a `ulong` key and failing due to it consistently returning null as the `UInt64` converted to `double` wouldn't ever match the key, because as it's documented in https://github.com/mbdavid/LiteDB/wiki/Data-Structure#json-extended, the `Int64` type is stored as a string value. 

This is the operator I'm talking about:
```cs
public static implicit operator BsonValue(UInt64 value)
{
    return new BsonValue((Double)value);
}
```
(https://github.com/mbdavid/LiteDB/blob/master/LiteDB/Document/BsonValue.cs#L431-L434)

pie-flavor:
If I had to guess, it'd be because `ulong.MaxValue` > `long.MaxValue`, so `double` can fit all values of `ulong` whereas `long` can't.

GreemDev:
@mbdavid Any info on this? I currently can't use LiteDB in my project because of InvalidCastExceptions.

mbdavid:
Hi guys, I will take a look on this.

GreemDev:
Just if you're interested, [this is the exception I'm getting.](https://img.greemdev.net/aivIVaHdad/VsDebugConsole_2019-11-20_00-56-40.png)
Int32 must be used internally, as I do not use Int32 anywhere for my database, only UInt64. 
In any event, there's the stacktrace. I honestly have no idea why this is happening, it never did before.

JKamsker:
Consolidated into #2898, which combines this report with the related ones, confirms it on current dev and tracks the fix. Closing this as a duplicate — please follow #2898, and comment there if your case is different.

PR #1227, opened 2019-05-20
Improved unsigned values support in `BsonValue`

falahati:
This PR adds support for `UInt32` values by saving them as an `Int64` value as well as improving `UInt64` support. Changes:

* Added a constructor for `BsonValue` accepting `UInt32` as an argument and saving it as an `Int64` value
* Added a `AsUInt32` property
* Implicit operators to convert from and to `UInt32`
* Added a constructor for `BsonValue` accepting `UInt64` as an argument and saving it as a `Double` value
* Added the missing `AsUInt64` property

### DIFF
--- a/LiteDB/Document/BsonValue.cs
+++ b/LiteDB/Document/BsonValue.cs
@@ -61,6 +61,10 @@ namespace LiteDB
             this.Type = BsonType.Int64;
             this.RawValue = value;
         }
+        
+        public BsonValue(UInt32 value) : this((Int64)value)
+        {
+        }
 
         public BsonValue(Double value)
         {
@@ -260,6 +264,11 @@ namespace LiteDB
             get { return this.IsNumber ? Convert.ToInt64(this.RawValue) : default(Int64); }
         }
 
+        public uint AsUInt32
+        {
+            get { return this.IsNumber ? Convert.ToUInt32(this.RawValue) : default(UInt32); }
+        }
+
         public double AsDouble
         {
             get { return this.IsNumber ? Convert.ToDouble(this.RawValue) : default(Double); }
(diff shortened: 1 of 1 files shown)


Emzi0767:
> Added a constructor for BsonValue accepting UInt64 as an argument and saving it as a Double value

Precision loss. This is not an enhancement. 64-bit integers should be stored as strings.

> Added a constructor for BsonValue accepting UInt32 as an argument and saving it as an Int64 value

U/Int32 should be treated as a 32-bit integer.

falahati:
> Precision loss. This is not an enhancement. 64-bit integers should be stored as strings.

Yeah, I am not happy with that personally. `BsonValue` should support `UInt32` internally. However, this was what @mbdavid decided to go for. I just improved the code by adding a constructor and a property to get the value. The decision to go with `double` was his. That being said `double` should be able in theory to represent `-Int64` to `+Int64` with no base change artifact. I need to double check this tho.

> U/Int32 should be treated as a 32-bit integer.
`UInt32` could be converted to and from `Int32` by using `unchecked` keyword but I decided to go with `Int64` due to the fact that this is supported by the `BsonValue` internally and have enough space to fit all valid values. The only downside is that it takes more space comparing to `UInt32`. However, it prevents clueless apps to read an invalid value for a large `UInt32` number. This includes LiteDB.Shell and any other app like my LiteDBViewer which don't have access to the model that the collection is created from.

Now to better solve both these problems, I think we should really add support for `UInt64` and `UInt32` directly into the `BsonValue` class. But for now, these are the best we can go for.

falahati:
Ok; base on this (if it is accurate):
https://en.wikipedia.org/wiki/IEEE_754

And especially this chart:
image

We should hit 10^0 imprecision (which is in decimal double that can make rounding problems) in for values bigger than about 10^16 which is less than ~10^19 that can be represented by `UInt64`. So double is officially a bad choice.

@mbdavid, what should be done to properly support `UInt64` internally? I might be able to do so if you point me in the right direction.

JKamsker:
Thanks for pushing on unsigned support, and for working through the precision question in the thread! Storing UInt64 as Double loses precision above 2^53 and doesn't match how the mapper stores ulong (as Int64). The fix for this is now being done in #2751 for #1224, so I'm closing this PR in favour of it.

PR #1256, opened 2019-07-23
Add AddRange<TCollection>(TCollection collection) To BsonArray

blowin:
initialize list with capacity


mbdavid:
Hi @blowin, thanks for your contribution. AddRange 

I saw that you remove some `this.` - I know that is optional but I used in all code for methods/properties in classes.

Also, `if` statement (or any block code) I always use single line or use `{`.

This king of patterns I will try to write in editorconfig to be default for every contribution.

blowin:
Ok. i understand it. I fixed code code convention

mbdavid:
Thanks @blowin!

PR #1257, opened 2019-07-24
Add ArrayPool

blowin:
Implement BufferPool


mbdavid:
Hi @blowin, thanks for this array pool implementation. I was avoiding use .net implementation because need add more libs. I will accept your PR and will review where I'm using this buffer pool (I read about renting arrays performance on small array sizes)

blowin:
Yes, I understood that you do not want to be attached to libraries. Therefore, made the implementation of the lightweight ArrayPool.

PR #1258, opened 2019-07-25
Problem with ArrayPool

blowin:
There is one problem in using ArrayPool. You borrow an array and save it in BufferSlice and then return it to ArrayPool, which can lead to data overwriting problems.

It may be worthwhile to create a wrapper over the array from the pool, make it Disposable, and also make BufferSlice Disposable and, when calling Dispose method, return the buffer back to the pool.


JensSchadron:
@blowin It seems like your PR got merge conflicts in the meantime, could you solve them when you've time?

blowin:
Yes of course. I fixed the conflicts.

PR #1276, opened 2019-08-08
Performance improvement in serializer.

blowin:
BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17134.885 (1803/April2018Update/Redstone4)
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
Frequency=2742192 Hz, Resolution=364.6718 ns, Timer=TSC
  [Host]    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.8.3801.0
  RyuJitX64 : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.8.3801.0

Job=RyuJitX64  Jit=RyuJit  Platform=X64

|       Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
| OldSerialize | 831.1 ns | 16.65 ns | 21.65 ns | 0.2546 |     - |     - |     802 B |
| NewSerialize | 736.7 ns | 14.37 ns | 21.06 ns | 0.2317 |     - |     - |     730 B |


LiteDB/Engine/Pages/BasePage.cs
```diff
@@ -225,7 +225,6 @@ public void MarkAsEmtpy()
             this.IsDirty = true;
 
             this.PageType = PageType.Empty;
-            this.ItemsCount = 0;
```
mbdavid: Why this line was removed?

LiteDB/Engine/Disk/Serializer/BufferPool.cs
```diff
@@ -35,5 +35,17 @@ public static void Return(byte[] buffer)
                 _bytePool.Return(buffer);
             }
         }
+
+        public static byte[] RentPageSizeBuff()
```
mbdavid: If will return `byte[]` why not use direct `new byte[]` in code? This use is very limited and checked if are not inside a loop or heavy use

LiteDB/Engine/Pages/BasePage.cs
blowin: This is old changes. You have the same line repeated twice https://github.com/blowin/LiteDB/blob/6c9d53a615216d615ada61ede7b7b9ab8f90ab4e/LiteDB/Engine/Pages/BasePage.cs#L151

LiteDB/Engine/Disk/Serializer/BufferPool.cs
blowin: I add this method, if in future improve BufferSlice. I wrote about this in comment:
https://github.com/mbdavid/LiteDB/pull/1258#issue-300993365

LiteDB.sln
```diff
@@ -1,45 +0,0 @@
-﻿
```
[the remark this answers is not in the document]
blowin: Of course. This is my mistake.

blowin:
This is may be useful for cases. when often create new small arrays. There will be less memory allocated, which will speed up garbage collection.

It is generally difficult to measure real performance problems in garbage collection. Since they occur while the application is running, they can be critical in certain situations.

PR #1339, opened 2019-10-11
Fix NullReferenceException in BsonMapper.ToDocument<T>()

### LINKED ISSUE #1338: BsonMapper.ToDocument<T>() can throw NullReferenceException
`BsonMapper.ToDocument<T>(object)` will throw a `NullReferenceException` when mapping an object which cannot be mapped to a `BsonDocument`, e.g: an array. However the non-generic version `BsonMapper.ToDocument(object)` will not throw, but return null instead.

Repro:
```csharp
var mapper = new BsonMapper();
var array = new int[] { 1, 2, 3, 4 };
var doc1 = mapper.ToDocument(typeof(int[]), array); // doc1 == null.
var doc2 = mapper.ToDocument(array); // NRE thrown.
```

FICTURE7:
Fixes #1338

### DIFF
--- a/LiteDB.Tests/Mapper/Mapper_Tests.cs
+++ b/LiteDB.Tests/Mapper/Mapper_Tests.cs
@@ -1,0 +1,21 @@
+﻿using FluentAssertions;
+using Xunit;
+
+namespace LiteDB.Tests.Mapper
+{
+    public class Mapper_Tests
+    {
+        private BsonMapper _mapper = new BsonMapper();
+
+        [Fact]
+        public void ToDocument_ReturnsNull_WhenFail()
+        {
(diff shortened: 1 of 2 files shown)


LiteDB.Tests/Mapper/Mapper_Tests.cs
```diff
@@ -0,0 +1,21 @@
...
+using Xunit;
+
+namespace LiteDB.Tests.Mapper
+{
+    public class Mapper_Tests
+    {
+        private BsonMapper _mapper = new BsonMapper();
+
+        [Fact]
+        public void ToDocument_ReturnsNull_WhenFail()
```
FICTURE7: Feel free to suggest me better names for this or if we even need this test. :>

Issue #1345, opened 2019-10-16
Strange memory usage


mregni:
Hi,

I'm working on a project and switch from SQLite to LiteDB and I love your project!

I do have a question about memory usage. It seems that when I query some data (database itself is 7Mb, 1100 documents) it does seems to take a lot of memory. 

Here is my code I'm using for this test to make sure it wasn't something else in my program.

```
int result;
var dbPath = Path.Combine(Directory.GetCurrentDirectory(), "data.db");
using (var database = new LiteDatabase($"FileName={dbPath}; journal=false"))
{
    var collection = database.GetCollection<Movie>();
    var list = collection.FindAll();
    result = list
        .SelectMany(x => x.Genres)
        .Distinct()
        .Count();
}

Console.WriteLine(result);
Console.ReadKey();
```

The result in my console is 20 (as expected) but it takes up to 90Mb in memory to calculate this even when the database itself is only 7Mb. Also, when this code arrives at the Console.ReadKey line, the memory usage is still 90Mb. It doesn't drop when going out of the using statement. 

This is only a small piece of my code, meaning that when I run a lot of queries it will add up upto 2Gb's of RAM. And when I rerun the job it adds up again to 4Gb, 6Gb, ... 

You have any idea what may cause this memory usage and how I can fix it?

Ow yeah, here is a screenshot of the memory usage in VS. The rise in memory happens on line 7
![image](https://user-images.githubusercontent.com/22617019/66958797-4ae0f580-f069-11e9-8196-93519f489555.png)

I also looked at your v5 alpha version, seems very promising because you can select only the properties you need and aggregations :) But I'm not sure I can start using it already.

FICTURE7:
#1351 might fix this.

I suspect that `PageBuffer` objects are surviving for longer than they should and since they maintain a reference to a possibly large `byte[]`, it too will survive for longer than it should as well. Therefore this causes that `byte[]` to remain uncollected after at least 1 garbage collections after it is has been marked for collection.

So if you did:
```csharp
using (var db = new LiteDatabase("file.db")) {
   // Work involving db.
}

GC.Collect(); // `PageBuffer` objects' finalizer must be executed before being collected -> promoted to higher generation.
```
Not a lot of memory would be reclaimed (assuming the largest allocations were done by LiteDB). However if you did:
```csharp
using (var db = new LiteDatabase("file.db")) {
   // Work involving db.
}

[...]

PR #1351, opened 2019-10-20
Remove unnecessary finalizers

FICTURE7:
Remove unnecessary finalizers which would end up in the finalizer queue, doing unnecessary work.

Objects with finalizers will survive at least one garbage collection and get promoted to a higher generation since the GC has to wait for their finalizers to get executed before reclaiming their memory. This means that they need at least 2 garbage collection before their memory is reclaimed which is especially bad for `PageBuffer` since it maintains a reference to a large `byte[]` object.

I did a simple test on a relatively large dataset (same dataset generated by LiteDB.Demo, that is, 1 million documents) and ran `col.FindAll().Count()` on it. It registered 29001 `PageBuffer` objects in the finalizer queue which is unnecessary, since `PageBuffer` does not do any meaningful work in its finalizer.

Before PR:
29001 `PageBuffer` objects registered and some other unnecessary ones by LiteDB.
Before

After PR:
After

Potential fix for #1345.

### DIFF
--- a/LiteDB/Client/Shared/SharedDataReader.cs
+++ b/LiteDB/Client/Shared/SharedDataReader.cs
@@ -29,11 +29,6 @@ namespace LiteDB
             _dispose();
         }
 
-        ~SharedDataReader()
-        {
-            this.Dispose();
-        }
-
         public BsonValue this[string field] => _reader[field];
 
         public string Collection => _reader.Collection;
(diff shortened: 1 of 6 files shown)


mbdavid:
Hi @FICTURE7, thanks for this great job. I saw you tests and is amazing. The only #if you keep it right: it's for debug proposes (on PageBuffer)

Thanks again!

PR #1381, opened 2019-12-07
Fix corrupt database when saving is interrupted

omfgicbf:
This fixes the order of writes and adds appropriate flushes when committing to disk.

This ensures that the journal is written to disk before proceeding with any changes and avoids leaving the database in an inconsistent state in case of a power failure.

### DIFF
--- a/LiteDB/Engine/Services/TransactionService.cs
+++ b/LiteDB/Engine/Services/TransactionService.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -75,6 +75,9 @@ namespace LiteDB
 
                 // mark header as recovery before start writing (in journal, must keep recovery = false)
                 header.Recovery = true;
+
+                // flush to disk to ensure journal is committed to disk before proceeding
+                _disk.Flush();
             }
             else
             {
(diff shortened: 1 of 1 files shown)


LiteDB/Engine/Services/TransactionService.cs
```diff
@@ -75,17 +75,32 @@ public void PersistDirtyPages()
 
                 // mark header as recovery before start writing (in journal, must keep recovery = false)
                 header.Recovery = true;
+
+                // flush to disk to ensure journal is committed to disk before proceeding
+                _disk.Flush();
```
mbdavid: `WriteJournal` method contains a a `Flush` call at end:

https://github.com/mbdavid/LiteDB/blob/v4/LiteDB/Engine/Disks/FileDiskService.cs#L194

omfgicbf: Yes, but `header.Recovery = true` is lost if the write is interrupted after that line. As there is no guarantee to the order of writes, if the changes below are committed but this is not, then you end up with a corrupt database.

mbdavid: You right, I miss this. I will accept your PR, thanks. New 4.1.5 will be released soon.

AviadKatani: Hi, if I already have a corrupted database from power failure, how can I fix it?

PR #1396, opened 2020-01-02
NRE fixes

KSemenenko:
Hi, I looked at your project code out of interest and found a couple of warnings and decided to fix it.


LiteDB/Document/Json/JsonWriter.cs
```diff
@@ -137,7 +137,7 @@ private void WriteArray(BsonArray arr)
                 var item = arr[i];
 
                 // do not do this tests if is not pretty format - to better performance
-                if (this.Pretty)
+                if (this.Pretty && item != null)
```
KSemenenko: item can be null

LiteDB/Document/Json/JsonWriter.cs
```diff
@@ -237,7 +237,7 @@ private void WriteKeyValue(string key, BsonValue value, bool comma)
             {
                 _writer.Write(' ');
 
-                if ((value.IsDocument && value.AsDocument.Keys.Any()) || (value.IsArray && value.AsArray.Count > 0))
+                if (value != null && ((value.IsDocument && value.AsDocument.Keys.Any()) || (value.IsArray && value.AsArray.Count > 0)))
```
KSemenenko: value can be null

LiteDB/Engine/FileReader/FileReaderV7.cs
```diff
@@ -112,7 +112,7 @@ public IEnumerable<BsonDocument> GetDocuments(string collection)
 
                         if (dataPage["pageType"].AsInt32 != 4) continue;
 
-                        var block = dataPage["blocks"].AsArray.FirstOrDefault(x => x["index"] == dataBlock["index"]).AsDocument;
+                        var block = dataPage["blocks"].AsArray.FirstOrDefault(x => x["index"] == dataBlock["index"])?.AsDocument;
```
KSemenenko: FirstOrDefault - can return null

LiteDB/Engine/Pages/BasePage.cs
```diff
@@ -233,7 +233,6 @@ public void MarkAsEmtpy()
             this.IsConfirmed = false;
 
             // block information
-            this.ItemsCount = 0;
```
KSemenenko: dublicate

LiteDB/Engine/FileReader/FileReaderV7.cs
JensSchadron: Not sure if it's actually possible to return "null" here. But in that case, it might be faster to use .First() instead of .FirstOrDefault()

LiteDB/Engine/Pages/BasePage.cs
JensSchadron: This has been fixed in the meantime by removing the other ItemCount assignment. If you remove this one as well, we'll be introducing a bug. Could undo this change?

KSemenenko: done

LiteDB/Engine/FileReader/FileReaderV7.cs
KSemenenko: next line will be
``` 
if (block == null) continue;
```
so we can return null, and as I checked no big differents betwen First and FirstOrDefault
https://github.com/microsoft/referencesource/blob/master/System.Core/System/Linq/Enumerable.cs
FirstOrDefault will retrn null, and First just throw an exeption

PR #1413, opened 2020-01-24
Initial setup for benchmark project

JensSchadron:
This is just the initial setup for the benchmarking project and includes currently only benchmarks for my testdata generator.

More benchmarks are currently being worked on, but it will take a bit more time.


mbdavid:
Hi @JensSchadron, your idea in benchmark is compare with v4? or compare with others databases? or only test performance?

JensSchadron:
Currently, my idea is to measure performance of LiteDB in general with different factors (like: encrypted or not, dataset size, simulteanous interactions with the database, et cetera), but I'm also thinking of benchmarking various parts of LiteDB (eg: the (de)serializer and such) so it might be easier to pinpoint which parts could use some further performance optimizations.

PR #1443, opened 2020-02-05
Feature/pragma constants

JensSchadron:
Better late than never, but here's the PR to add a class which contains the constants for the pragmas. However, I'm currently not sure whether the name of the static class (and the location of the same class) are good.
Let me know whether I need to move/rename the constants class.


mbdavid:
Thanks Jens! I will just add XML comments in each pragma constant

PR #1446, opened 2020-02-06
Optimize performance of IsAnonymousType method.

yaho86:
I have came across performance drop when reading from DB after upgrade from v3 to v5.
The profiler shows that one of bottlenecks is `IsAnonymousType` method which is quite expensive.

This fix resolves two problems.
1. call `IsAnonymousType` only if `BsonValue.IsDocument`
2. optimise `IsAnonymousType` method - use `Attribute.IsDefined` instead of `GetCustomAttributes`, because it is more performant (there is `#if` because it is not available for .NET Standard 1.3)

### DIFF
--- a/LiteDB/Client/Mapper/BsonMapper.Deserialize.cs
+++ b/LiteDB/Client/Mapper/BsonMapper.Deserialize.cs
@@ -135,12 +135,6 @@ namespace LiteDB
                 return custom(value);
             }
 
-            // if type is anonymous use special handler
-            else if(type.IsAnonymousType() && value.IsDocument)
-            {
-                return this.DeserializeAnonymousType(type, value.AsDocument);
-            }
-
             // if value is array, deserialize as array
             else if (value.IsArray)
             {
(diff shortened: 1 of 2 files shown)


Tirraon:
Thanks!
I don't know how slow is Attribute.IsDefined or GetCustomAttributes, but I think it should check FullName at first as faster condition and check for attribute only if name is correct.

yaho86:
Very good point!

`FullName` check is much faster than Attribute check. Maybe then it is not necessary to use `Attribute.IsDefined` and avoid #if (always use `GetCustomAttributes`) to prefer simpler code.

I will update the PR.

yaho86:
There are some numbers to support these changes. 

Reading 1 000 000 items from DB (single collection).
Items are represented by this simple class
```c#
class Entity 
{
    public object Data { get; set; }
}
```

Before optimization: **3443** ms
After optimization: **1888** ms

Performance improvement is most noticeable when using `object` type properties in DB entities.

JensSchadron:
@yaho86 I'll try to create a Benchmark for this case as well 🙂

yaho86:
@JensSchadron I used just simple ``FindAll()`` to get these numbers. This optimization helps always if deserialized class has properties of type ``object``.

Anyway, thanks for the merge and great work with LiteDB 🙂

PR #1480, opened 2020-02-15
Fix Deserialize Object (as Dictionary) and Hashtable.

nightroman:
Problem: BsonMapper correctly creates Dictionary and Hashtable instances but the collections are empty.
P.S. Hashtables are important, for example they are used as "native dictionaries" in PowerShell.

The original line `if (o is IDictionary && typeInfo.IsGenericType)` has two problems:

- it excludes non-generic dictionaries, for example `Hashtable` (quite important type in PowerShell)
- it fails to deserialize `object` as `Dictionary` because it checks `typeInfo` obtained for `object` (non-generic of course)


mbdavid:
Thans @nightroman.

PR #1491, opened 2020-02-19
Use thread pool in DiskWriterQueue

vuplea:
https://github.com/mbdavid/LiteDB/issues/1490


mbdavid:
Thanks @vuplea, this code was a PR from old implementation. Thanks for this.

Issue #1506, opened 2020-02-23
[SOLVED but...] Query.Limit is ignored by Find(Query)


nightroman:
UPDATE: The problem seems to be in the unusual API design, see next comments.

**Version**

LiteDB 5.0.3 and the latest from repo, Windows 10, net45

**Describe the bug**

`Query.Limit` is ignored by collection `Find(Query)`.

**Code to Reproduce**

```csharp
[Fact]
public void Query_Limit()
{
    using (var db = new LiteDatabase(":memory:"))
    {
        var col = db.GetCollection("Test");
        col.Insert(new BsonDocument { ["_id"] = 1 });
        col.Insert(new BsonDocument { ["_id"] = 2 });
        col.Count().Should().Be(2);

        var query = new Query();
        query.Limit = 1;
        var r = col.Find(query).ToList();
        r.Count.Should().Be(1); // ~ fails, actual is 2
    }
}
```

**Expected behavior**

The above test is expected to pass, `Query.Limit` set to 1 should result in 1 result document, not actual 2.

nightroman:
And a side note (let me know if it's worth another issue to open).

From ILSpy, this the code of another `Find(Query, int, int)`:

```csharp
public IEnumerable<T> Find(Query query, int skip = 0, int limit = int.MaxValue)
{
    if (query == null)
    {
        throw new ArgumentNullException("query");
    }
    query.Offset = skip;
    query.Limit = limit;
    return new LiteQueryable<T>(_engine, _mapper, _collection, query).ToEnumerable();
}
```

This code changes the input `query`. This does not look quite right.
The `query` may be used later in `Find(Query)` with unexpectedly changed `Limit` and `Offset`.

nightroman:
Ah, I get it... The two issues are in fact one. `Find` with default skip and limit parameters overrides my `Offset` and `Limit` settings in the passed `query`.

Is this design deliberate and thought through?
If yes, please close the issue.

lbnascimento:
@nightroman, thank you for your question. This behavior is not intended and is fixed in the master branch. It should be present in the next incremental release.

nightroman:
Thank you.

```
-            query.Offset = skip;
-            query.Limit = limit;
+            if(skip != 0) query.Offset = skip;
+            if(limit != int.MaxValue) query.Limit = limit;
```

It's still does not look quite right (to me). Because non-default parameters still change the input `query` and if this `query` is used afterwards on it's own this change may be unexpected.

[...]

PR #1514, opened 2020-02-27
Fix limit and offset Query parameters being overriden with default values in Find method call

lbnascimento:
See #1506

### DIFF
--- a/LiteDB/Client/Database/Collections/Find.cs
+++ b/LiteDB/Client/Database/Collections/Find.cs
@@ -41,8 +41,8 @@ namespace LiteDB
         {
             if (query == null) throw new ArgumentNullException(nameof(query));
 
-            query.Offset = skip;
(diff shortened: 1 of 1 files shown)


LiteDB/Client/Database/Collections/Find.cs
```diff
@@ -41,8 +41,8 @@ public IEnumerable<T> Find(Query query, int skip = 0, int limit = int.MaxValue)
         {
             if (query == null) throw new ArgumentNullException(nameof(query));
 
-            query.Offset = skip;
-            query.Limit = limit;
+            if(skip != 0) query.Offset = skip;
+            if(limit != int.MaxValue) query.Limit = limit;
```
nightroman: Please see my comment https://github.com/mbdavid/LiteDB/issues/1506#issuecomment-591949738

PR #1456, opened 2020-02-08
Initial multithreaded insertion benchmarks

JensSchadron:
Hi @mbdavid 
So... I've finally found some time to finish up the multithreaded insertion benchmarks.

I've ran the benchmarks on my own machine and this is the summary.
Please note that something strange is going on related to the mean runtimes, that I can't explain for now, when running against dataset sizes of 10k items and 20 simultaneous threads. [You might want to inspect the "detailed" logs as well when you've time to investigate this](https://github.com/mbdavid/LiteDB/files/4175475/LiteDB.Benchmarks.Benchmarks.Insertion.MultiThreadedInsertionBenchmark-20200207-230753.log)

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
AMD Ryzen 9 3900X, 1 CPU, 24 logical and 12 physical cores
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
  Job-BZNOKG : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT

Jit=RyuJit  Runtime=.NET Core 3.1  Force=True  
Toolchain=.NET Core 3.1  InvocationCount=1  UnrollFactor=1  

|            Method | DatasetSize |       Password | AmountOfThreads |            Mean |           Error |          StdDev |          Median |         Gen 0 |      Gen 1 |     Gen 2 |       Allocated |
|------------------ |------------ |--------------- |---------------- |----------------:|----------------:|----------------:|----------------:|--------------:|-----------:|----------:|----------------:|
| InsertionYThreads |          10 |              ? |               1 |        205.6 us |         4.08 us |         7.77 us |        203.7 us |             - |          - |         - |         61.3 KB |
| InsertionYThreads |          10 |              ? |               2 |        242.6 us |         4.85 us |        13.84 us |        239.8 us |             - |          - |         - |       122.99 KB |
| InsertionYThreads |          10 |              ? |               5 |        299.3 us |         7.17 us |        20.70 us |        296.2 us |             - |          - |         - |       319.87 KB |
| InsertionYThreads |          10 |              ? |              10 |        423.4 us |        13.12 us |        36.79 us |        419.5 us |             - |          - |         - |       630.82 KB |
[...]

Issue #1521, opened 2020-02-29
[BUG] .Find() shows Object reference not set to an instance of an object.


josephinenewbie:
**Version**
Upgraded to v5.0.3 from v4.1.4

**Describe the bug**
When try to .Find() with query inside, it shows Object reference not set to an instance of an object.

**Code to Reproduce**
```
        public IEnumerable<Member> GetActiveMembers()
        {
            return databaseInstance.GetCollection<Member>("Members")    
                    .Find(Query.Or(
                            Query.GTE("ExpireOnDateTime", DateTime.Now),
                            Query.EQ("ExpireOnDateTime", null)
                          ))
                    .OrderBy(x => x.Name);
        }
```

**Expected behavior**
Previously above code works find in v4.1.3. 

**Screenshots/Stacktrace**
```
An unhandled exception has occurred while executing the request.
System.NullReferenceException: Object reference not set to an instance of an object.
   at LiteDB.BsonExpressionParser.TryParseFunction(Tokenizer tokenizer, ExpressionContext context, Boolean isRoot)
   at LiteDB.BsonExpressionParser.ParseSingleExpression(Tokenizer tokenizer, ExpressionContext context, Boolean isRoot)
   at LiteDB.BsonExpressionParser.ParseFullExpression(Tokenizer tokenizer, ExpressionContext context, Boolean isRoot)
   at LiteDB.BsonExpression.Parse(Tokenizer tokenizer, BsonExpressionParserMode mode, Boolean isRoot)
   at LiteDB.BsonExpression.Create(String expression)
   at LiteDB.Query.EQ(String field, BsonValue value)
```

Thanks again.

JensSchadron:
I stumbled upon this issue myself as well earlier today during the benchmarks. Still looking into it to see what's causing it because creating the BsonExpression manually does seem to work. 🤔 

For now, I've pinned the issue down to the Query.Or() BsonExpression, but I'll look into it further tomorrow.

mbdavid:
Hi guys, the bug is on `Query.EQ("...", **null**))`. I'm fixing. For now, use `BsonValue.Null`

JensSchadron:
@josephinenewbie, just wanted to let you know that this issue (and the one I referenced #1527 ) will be fixed in the next update and therefore, I'll be closing this ticket.

If you come across a thsis issue after you've installed to new version (when it's available), please open a new issue (or reopen this ticket if you think it's exactly the same) 🙂

Issue #1478, opened 2020-02-14
Delete all items from collection


GW-FUB:
With LiteDB v4 is was used to write `Delete(Query.All())`. This doesn't work with v5 anymore. Also using `DeleteMany(_ => true)` gets me an exception.

How can I delete all entries from a collection?

JensSchadron:
@GW-FUB Not completely sure how it's supposed to be, but this does the job 🙂
```csharp
class Program
{
	static void Main(string[] args)
	{
		using var db = new LiteDatabase(new MemoryStream());
		var col = db.GetCollection<TestClass>();
		col.Insert(new TestClass());
		col.Insert(new TestClass());

		Console.WriteLine($"{nameof(TestClass)} collection contains {col.Count()} item(s)\n");

		Console.WriteLine($"Deleting item(s) in collection {nameof(TestClass)}\n");
		col.DeleteMany("1=1");

		Console.WriteLine($"{nameof(TestClass)} collection contains {col.Count()} item(s)\n");
	}

	class TestClass
	{
		public ObjectId Id { get; set; }
		public string Data { get; set; }
	}
}
```

mbdavid:
Hi @JensSchadron, I just add support for unary parameter as predicate in LINQ converter.

JensSchadron:
@mbdavid I tested it and it seems to work indeed. Therefor I'll close this issue.

@GW-FUB Not sure if you're still interested in this, but you can now also pass `_ => true` to the `DeleteMany()` method.

PR #1522, opened 2020-02-29
Feature/performance benchmarks v5

JensSchadron:
Hereby a set of basic benchmarks related to insertion/upsertion, querying and deletion
My desktop is still running the benchmarks so I'll post the results later when it's done.
If you notice any benchmarks that have been set up incorrectly (or are missing), please let me know. Or feel free to add them later yourself.

As soon as this PR has been merged, I'll do some changes in #1456 so that the benchmarks have been set up in the same way.

### DIFF
--- a/LiteDB.Benchmarks/Benchmarks/BenchmarkBase.cs
+++ b/LiteDB.Benchmarks/Benchmarks/BenchmarkBase.cs
@@ -1,0 +1,31 @@
+using BenchmarkDotNet.Attributes;
+
+namespace LiteDB.Benchmarks.Benchmarks
+{
+    public abstract class BenchmarkBase
+    {
+        // Insertion data size
+        [Params(10, 50, 100, 500, 1000, 5000, 10000)]
+        public int DatasetSize;
+
+        public virtual string DatabasePath
+        {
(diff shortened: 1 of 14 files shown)


JensSchadron:
Seems like a few benchmarks are broken here and there, will take a bit more time to fix them 😅

JensSchadron:
@mbdavid The 2 issues I stumbled upon are the following ones
- Lambda expressions which just return a constant value throwing an exception (eg: _ => true) #1478 
- Query.Or() (and possibly Query.And() ) also throwing an exception #1521

I've been looking into both of them, but haven't found a solution yet. So if you (or someone else) could help me out on this, that would be great. 🙂

mbdavid:
Hi @JensSchadron, should I wait before merge next PR #1456? You have permissions in repo to merge your benchmarks tests :)

JensSchadron:
@mbdavid I know I do have the "permission" to merge into the repo directly, however... I think it would be more than useful if people could review the code that I (or others) write so possible bugs/issues can be detected before they make their way into the repo.

I'll also do a few more changes in PR #1456 so please hold out a bit longer before reviewing the code. I'll notify when it's ready. 🙂

Issue #1544, opened 2020-03-09
[BUG] Query() Return only 1 record if execute Count() before execute ToList()


leung85:
**Version**
5.0.3

**Describe the bug**
Run  _col.Query().ToList() directly will have 20000 records
But return only 1 record if use same query and execute Count(); before execute  ToList();

**Code to Reproduce**
```
            var query = _col.Query();
            int totalCount = query.Count() ; // 20000 record
            var result = query.ToList(); // only 1 record
            return (result, totalCount);
```

**Expected behavior**
Return 20000 record 

**Additional context**
What i want to do is to get the total count after filtering, and then use Limit(size).Offset(offset).toList(); for paging.

lbnascimento:
@leung85 This issue was fixed in the master branch and will be present in the next incremental release.

PR #1549, opened 2020-03-09
Fix LiteQueryable.Count() overwriting query

lbnascimento:
See #1544


nightroman:
Are there any exceptions possible between storing and restoring the query value?
If yes then this approach might be still fragile.

lbnascimento:
@nightroman You are correct, I'll add try-finally blocks

PR #1562, opened 2020-03-13
Add missed throw keyword

maxkatz6:
Between 5.0.2 and 5.0.3 exception was changed on these lines, but "throw" keyword was missed.

It can cause unexpected behavior, if page is parsed with invalid PageType, what is possible in some cases like in https://github.com/mbdavid/LiteDB/issues/1503


mbdavid:
Thanks for fix @maxkatz6 !

PR #1602, opened 2020-03-30
Added an IsDisposed method to LiteDatabase class.

andremorata:
I created this property because currently we do not have a way to check if an existing instance is still valid.

### DIFF
--- a/LiteDB/Client/Database/LiteDatabase.cs
+++ b/LiteDB/Client/Database/LiteDatabase.cs
@@ -24,6 +24,11 @@ namespace LiteDB
         /// </summary>
         public BsonMapper Mapper => _mapper;
 
+        /// <summary>
+        /// Returns true if the current instance is already disposed
+        /// </summary>
+        public bool IsDisposed { get; set; } = false;
+
         #endregion
 
         #region Ctor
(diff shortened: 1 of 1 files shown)


nightroman:
This is an unusual pattern... Why just set a reference to null, as soon as it is disposed (more used pattern)?

andremorata:
> Out of curiosity, what is the use case? Why not to set a reference to null, as soon as it is disposed (typical pattern)? And instead of checking for this property check for null.

If you want to use a single point for the connection (inside a service for instance), it might be in a currently disposed state after an action performed inside an using statement, but that instance is still alive/valid after that so it does not have a null value. Having this property available you can perform the necessary checks and only instantiate a new one when is actually needed.

nightroman:
If this PR is accepted then this property should be added to the interface ILiteDatabase as well.

Another issue/question: if this approach is legit and needed then same properties are needed in similar types ILiteEngine, LiteEngine, SharedEngine, and LiteRepository. Otherwise introducing it in just database is going to be somewhat inconsistent. All these types may be used in similar scenarios.

fubar-coder:
TBH: I don't see a need for this change. You can simply write/use a wrapper class (which might even be derived from LiteDatabase) and this code doesn't really add any functionality to LiteDb itself. Continuing using a disposed object seems like a bad thing to do.

PR #1605, opened 2020-04-01
Make MAX_DEPTH a variable property

PietroBellone:
The limit of 20 may not me enough for some situations


DrTtnk:
Changes: 
 - made MAX_DEPTH a public property with default value 20

nightroman:
I also think that 20 is too small. I sometimes need to "dump" some unknown PowerShell custom objects for further analysis later and, from experience, the depth may exceed 20.

Just in case MongoDB has this limit set to 100.

It would be nice to have this limit configurable (ideally) or at least increased to 100, to be on par with MongoDB.

PR #1609, opened 2020-04-02
fix(cache): switched to concurrent cache

### LINKED ISSUE #1416: Concurrency issue while Querying
In LiteDB 5 it appears you cannot use the result of a query as an Enumerable to pass to Parallel.Foreach if you are performing additional query operations on the database inside of the Parallel.Foreach loop.

The error message received is "System.InvalidOperationException: 'Collection was modified; enumeration operation may not execute.'" this is thrown on `var exists = col.Exists(x => x.IdentityHash == IdentityHash && x.RunId == RunId);` inside `WriteObjectExists`

I'm not performing any writes while this is being run.  The below code can reproduce the issue with a few thousand objects in the database.  I've tried generating the IEnumerable different ways (as shown below)

    public static IEnumerable<WriteObject> GetMissingFromFirst2(string firstRunId, string secondRunId)
    {
        var col = db.GetCollection<WriteObject>("WriteObjects");

        var list = new ConcurrentBag<WriteObject>();

        var wos = col.Find(x => x.RunId == secondRunId);

        Parallel.ForEach(wos, wo =>
        {
            if (!WriteObjectExists(firstRunId, wo.IdentityHash))
            {
                list.Add(wo);
            }
        });

        return list;
    }

    public static IEnumerable<WriteObject> GetMissingFromFirst(string firstRunId, string secondRunId)
    {
        var bag= new ConcurrentBag<WriteObject>();

        var identityHashes = db.Execute($"SELECT IdentityHash FROM WriteObjects WHERE RunId = @0",
                new BsonDocument
                {
                    ["0"] = secondRunId
                });

        Parallel.ForEach(identityHashes.ToEnumerable(), IdentityHash =>
        {
            if (!WriteObjectExists(firstRunId, IdentityHash["IdentityHash"].AsString))
            {
                list.Add(GetWriteObject(secondRunId, IdentityHash.AsString));
            }
        });
        return bag;
    }

    private static bool WriteObjectExists(string RunId, string IdentityHash)
    {
        var col = db.GetCollection<WriteObject>("WriteObjects");
        var exists = col.Exists(x => x.IdentityHash == IdentityHash && x.RunId == RunId);
        return exists;
    }

    private static WriteObject GetWriteObject(string runId, string IdentityHash)
    {
        var col = db.GetCollection<WriteObject>("WriteObjects");
        return col.FindOne(x => x.IdentityHash == IdentityHash && x.RunId == runId);
    }

OneCyrus:
fixes #1416


LiteDB/Engine/Services/CacheService.cs
```diff
@@ -1,11 +1,18 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
+
+#if !NET35
```
maxkatz6, line 6: LiteDB doesn't support NET35, so these conditions could be removed here and below.

maxkatz6:
And it looks like lock in ClearPages is not needed anymore, because ConcurrentDicrionary has lock inside Clear() https://github.com/dotnet/runtime/blob/master/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L590

LiteDB/Engine/Services/CacheService.cs
OneCyrus: this fixes v4 which targets 3.5

OneCyrus:
@mbdavid any chance we get a bugfix release for 4.x with this issue resolved?

JKamsker:
Thanks for the contribution, and sorry it sat so long. This targets the v4 `CacheService`, which no longer exists: v5 replaced it with `MemoryCache`, and #2772 now rewrites that with a bounded, concurrency-safe cache. Closing, as it can no longer be applied.

PR #1639, opened 2020-04-17
Check if the read lock is held before exiting

huyhu:
Check if the read lock is held before exiting.

### DIFF
--- a/LiteDB.Tests/Engine/ParallelQuery_Tests.cs
+++ b/LiteDB.Tests/Engine/ParallelQuery_Tests.cs
@@ -10,7 +10,7 @@ namespace LiteDB.Tests.Engine
 {
     public class ParallelQuery_Tests
     {
-        [Fact(Skip = "Must fix parallel query fetch")]
+        [Fact]
         public void Query_Parallel()
         {
             using(var db = new LiteDatabase(new MemoryStream()))
--- a/LiteDB/Engine/Services/LockService.cs
+++ b/LiteDB/Engine/Services/LockService.cs
@@ -54,7 +54,8 @@ namespace LiteDB.Engine
             // if current thread are in reserved mode, do not exit transaction (will be exit from ExitExclusive)
             if (_transaction.IsWriteLockHeld) return;
 
-            _transaction.ExitReadLock();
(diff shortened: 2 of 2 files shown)


LiteDB.Tests/Engine/ParallelQuery_Tests.cs
```diff
@@ -10,7 +10,7 @@ namespace LiteDB.Tests.Engine
 {
     public class ParallelQuery_Tests
     {
-        [Fact(Skip = "Must fix parallel query fetch")]
```
huyhu, line 13: Voila! the parallel query works with this change.

LiteDB/Engine/Services/LockService.cs
```diff
@@ -54,7 +54,8 @@ public void ExitTransaction()
             // if current thread are in reserved mode, do not exit transaction (will be exit from ExitExclusive)
             if (_transaction.IsWriteLockHeld) return;
 
-            _transaction.ExitReadLock();
+            if (_transaction.IsReadLockHeld)
```
huyhu, line 57: Check before exiting, otherwise ExitReadLock throws System.Threading.SynchronizationLockException:The read lock is being released without being held

lbnascimento:
@huyhu Does this actually fix your use case? Do you have any tests for this PR? I'm worried this PR only masks an underlying issue (why did the transaction try to release a lock that it didn't hold in the first place?)

dethknite:
I would agree it is important to find the underlying cause of the issue, however, adding this check does stop a fatal exception that takes down a an application. I, too, have encountered this error several times of the past few months, and I have not been able to isolate it to anything in particular. (I am using the mutex locking scheme with async tasks).

Perhaps a if (held) {remove} / else {log} type of solution would suffice without reviewing all the locking logic and retaining this bug.

JKamsker:
Thanks for the fix! The same IsReadLockHeld guard landed in v5.0.17 and is on dev. As noted in review, it only hides the symptom: the lock is still held by the original thread. The underlying thread-affinity problem is tracked in #2790, so I'm closing this PR.

PR #1657, opened 2020-04-27
docs: fix hugo links

viceice:
The relative links at the docs are wrong [here](https://www.litedb.org/docs/indexes/)


viceice:
https://gohugo.io/content-management/cross-references/

viceice:
Does not work yet. Do i need to rebuild locally and send pr with updated files?

lbnascimento:
@viceice Yes, you need to rebuild after every change

viceice:
ok, so  I'll send a new pr with rebuild

Issue #1692, opened 2020-05-20
[SUGGESTION] I wish to have a feature for handling LiteDb with Isolated Storage


pankaj-nikam:
**Is your feature request related to a problem? Please describe.**
I wish to store some information where I believe `IsolatedStorage` is the best candidate. The dotnet API does not provide a full path for the `IsolatedStorage` file where we can pass it to LiteDb to store the database (with reflection its possible, however it is a hack and not a perfect solution I believe).

**Describe the solution you'd like**
I would like a constructor which accepts `IsolatedStorageFile` instance and with the other options like Password and other connection string options we can pass.

**Describe alternatives you've considered**
I tried it with the *hack* and was not satisfied with the flow of the code, for just getting a field, I had to use reflection, and it is internal which may get renamed anytime.

**Additional context**
I would be happy if there is another constructor overload with the options mentioned above.

quicksln:
Hi 

I think it is already possible with ```IsolatedStorageFileStream``` and ```LiteDatabase``` constructor overload type ```ILiteEngine``` implementation. 

I was able to ready LiteDB instance from isolate storage and read / insert documents. Below I’m attaching sample code that I think can be enhanced for your needs.

```csharp
    class Program
    {
        static void Main(string[] args)
        {
            IsolatedStorageFile isoFile = IsolatedStorageFile.GetUserStoreForAssembly();

            String[] dirNames = isoFile.GetDirectoryNames("*");
            if (dirNames.Contains("LiteDB") && isoFile.FileExists("LiteDb\\LiteDb.dll")) {
                IsolatedStorageFileStream liteDBStream = isoFile.OpenFile("LiteDb\\LiteDb.dll", System.IO.FileMode.Open);

                var connectionType = LiteDB.ConnectionType.Shared;

                EngineSettings engineSettings = new EngineSettings();
                engineSettings.DataStream = liteDBStream;
                engineSettings.Filename = "LiteDb.dll";
                engineSettings.Password = "1234";
                engineSettings.ReadOnly = false;

                ILiteEngine liteDbEngine = connectionType == LiteDB.ConnectionType.Shared ? 
                    (ILiteEngine)new LiteDB.SharedEngine(engineSettings) : (ILiteEngine)new LiteEngine(engineSettings);

                LiteDB.LiteDatabase liteDatabase = new LiteDB.LiteDatabase(liteDbEngine);

                var testQuery = liteDatabase.GetCollection<TestDto>("query");
[...]

PR #1703, opened 2020-06-01
Introducing new ctor paramter for LiteDatabase

quicksln:
Hi 
PR has been created for suggestion #1692. 

I’ve introduced new parameter to existing cotr and altered existing xunit tests. I’ve try to came up with new unit test idea that would test read/write scenario for LogStream, but it already exists in Disk_Tests.Disk_Read_Write()

Cheers
Piotr

### DIFF
--- a/LiteDB/Client/Database/LiteDatabase.cs
+++ b/LiteDB/Client/Database/LiteDatabase.cs
@@ -56,14 +56,17 @@ namespace LiteDB
         }
 
         /// <summary>
-        /// Starts LiteDB database using a generic Stream implementation (mostly MemoryStrem).
-        /// Use another MemoryStrem as LOG file.
+        /// Starts LiteDB database using a generic Stream implementation (mostly MemoryStream).
         /// </summary>
-        public LiteDatabase(Stream stream, BsonMapper mapper = null)
+        /// <param name="stream">DataStream reference </param>
+        /// <param name="mapper">BsonMapper mapper reference</param>
+        /// <param name="logStream">LogStream reference </param>
+        public LiteDatabase(Stream stream, BsonMapper mapper = null, Stream logStream = null)
         {
             var settings = new EngineSettings
             {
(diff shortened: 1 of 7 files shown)


LiteDB/Client/Database/LiteDatabase.cs
```diff
@@ -56,14 +56,17 @@ public LiteDatabase(ConnectionString connectionString, BsonMapper mapper = null)
...
+        /// <param name="stream">DataStream reference </param>
+        /// <param name="mapper">BsonMapper mapper reference</param>
+        /// <param name="logStream">LogStream reference </param>
+        public LiteDatabase(Stream stream, BsonMapper mapper = null, Stream logStream = null)
         {
             var settings = new EngineSettings
             {
-                DataStream = stream ?? throw new ArgumentNullException(nameof(stream))
+                DataStream = stream ?? throw new ArgumentNullException(nameof(stream)),
+                LogStream = logStream
```
lbnascimento, line 69: I think this breaks when `logStream` is null. Maybe you should make it `LogStream = logSteam ?? new MemoryStream()`

quicksln: Hi @lbnascimento thank you for feedback and code review.

Just double checked it and I think additional checking for log stream is not necessary at this point.  Please have a look on following code path call :

LiteEngine cotr , line 70 :
```csharp
                _disk = new DiskService(settings, MEMORY_SEGMENT_SIZES);
```
DiskService ctro,  line 37 :
```csharp
            _logFactory = settings.CreateLogFactory();
```
and CreateLogFactory method line 108 : 
```csharp
            return new StreamFactory(new MemoryStream(), this.Password);
```

In CreateLogFactory factory method if logStream is null StreamFactory will be instances with new MemoryStream()

Let me know what you think about it and if I didn't miss anything. Still learning LiteDB :)

Regards
Piotr

Issue #1860, opened 2020-10-27
constructor with enum parameter crashes


archaeron:
**Version**
5.0.9

**Describe the bug**
I have a class with a single, public constructor that takes an enum as a parameter.
The code crashes when it tries to get that class from the Database.
The same thing works without a public constructor by making all fields nullable.

**Code to Reproduce**
```csharp
#nullable enable

using System;
using System.Linq;
using LiteDB;

namespace litedb_example
{
    [Serializable]
    public enum EnumAB
    {
        [System.Runtime.Serialization.EnumMember(Value = @"a")]
        A,
        [System.Runtime.Serialization.EnumMember(Value = @"b")]
        B,
    }

    public class C1
    {
        public int Id { get; set; }

        public EnumAB? EnumAB {get; set; }
    }

    public class C2
    {
        public int Id { get; set; }

        public EnumAB EnumAB {get; set; }

        public C2(int id, EnumAB enumAB)
        {
            Id = id;
            EnumAB = enumAB;
        }
    }

    class Program
    {
        static void Main(string[] args)
        {

            // Open database (or create if doesn't exist)
            using (var db = new LiteDatabase(@"./example.db"))
            {
                // Get a collection (or create, if doesn't exist)
                var col1 = db.GetCollection<C1>("c1");
                var col2 = db.GetCollection<C2>("c2");

                var c1 = new C1
                {
                    Id = 1,
                    EnumAB = EnumAB.A
                };

                col1.Insert(c1);

                var c2 = new C2
                (
                    id: 1,
                    enumAB: EnumAB.A
                );

                col2.Insert(c2);

                var value1 = col1.FindAll().FirstOrDefault();
                System.Console.WriteLine(value1);

                var value2 = col2.FindAll().FirstOrDefault();
                System.Console.WriteLine(value2);
            }
        }
    }
}
```

**Expected behavior**
I would expect either both variants to fail or both to work.

**Screenshots/Stacktrace**
Unhandled exception. LiteDB.LiteException: Failed to create instance for type 'litedb_example.C2' from assembly 'litedb_example.C2, litedb-example, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'. Checks if the class has a public constructor with no parameters.
 ---> System.ArgumentException: Type 'litedb_example.C2' does not have a default constructor (Parameter 'type')
   at System.Linq.Expressions.Expression.New(Type type)
   at LiteDB.Reflection.CreateClass(Type type)
[...]

PR #1867, opened 2020-11-05
fix constructor with enum parameter crashes

for7raid:
fix issue about enums in constructor #1860


archaeron:
this is great, thanks @for7raid  !

this will fix quite a few things for me and comes right in time for C# 9 (where nullable will be much more of a thing) that is coming out at build tomorrow.

lbnascimento:
@for7raid Thank you for your contribution!

PR #1868, opened 2020-11-08
fix #1865 Wrong type after include directive

### LINKED ISSUE #1865: [BUG] Wrong type after include directive
**Version**
Core 3.1, LiteDB 5.0.9

**Describe the bug**
I have a base class and its descendants. Objects can refer to each other in the base class among themselves.
When I fetch the data, the types match those that were stored. But when I use Include, I get the base type, not real saved object.

**Code to Reproduce**
```
 public class Project : BaseEntity
        {
           
        }

        public class Point : BaseEntity
        {
            public BaseEntity Project { get; set; }
            public BaseEntity Parent { get; set; }
            public DateTime Start { get; internal set; }
            public DateTime End { get; internal set; }
        }

        public class BaseEntity
        {
            public ObjectId Id { get; set; } = ObjectId.NewObjectId();
        }

            BsonMapper.Global.Entity<Point>().DbRef(p => p.Project);
            BsonMapper.Global.Entity<Point>().DbRef(p => p.Parent);
            BsonMapper.Global.ResolveCollectionName = (s) => "activity";

            using var _database = new LiteDatabase(":memory:");

            var project = new Project() { Name = "Project" };
            var point1 = new Point { Parent = project, Project = project, Name = "Point 1", Start = DateTime.Now, End = DateTime.Now.AddDays(2) };
            var point2 = new Point { Parent = point1, Project = project, Name = "Point 2", Start = DateTime.Now, End = DateTime.Now.AddDays(2) };


            _database.GetCollection<Point>("activity").Insert(point1);
            _database.GetCollection<Point>("activity").Insert(point2);
            _database.GetCollection<Project>("activity").Insert(project);


            var p1 = _database.GetCollection<Point>()
                .FindById(point1.Id);
//The project and parent fields are real project type
            Assert.Equal(typeof(Project), p1.Parent.GetType()); 
            Assert.Equal(typeof(Project), p1.Project.GetType());

            p1 = _database.GetCollection<Point>()
                .Include(p=>p.Parent).Include(p=>p.Project)
                .FindById(point1.Id);
//Issue here - project and parent are BaseType
            Assert.Equal(typeof(Project), p1.Parent.GetType());
            Assert.Equal(typeof(Project), p1.Project.GetType());
```

**Expected behavior**
After deserialization the types of object are real types

**Additional info**
The `$type` field in bson document is ignored, when object force included

for7raid:
fix #1865

### DIFF
--- a/LiteDB.Tests/Issues/Issue1651_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue1651_Tests.cs
@@ -5,22 +5,24 @@ using System.Linq;
 
 namespace LiteDB.Tests.Issues
 {
-    public class Order : BaseEntity
-    {
-        public Customer Customer { get; set; }
-    }
-
-    public class Customer : BaseEntity
-    {
-        public string Name { get; set; }
-    }
-
-    public class BaseEntity
-    {
-        public Guid Id { get; set; }
-    }
+    
     public class Issue1651_Tests
     {
+        public class Order : BaseEntity
+        {
+            public Customer Customer { get; set; }
+        }
+
+        public class Customer : BaseEntity
+        {
+            public string Name { get; set; }
+        }
+
+        public class BaseEntity
+        {
+            public Guid Id { get; set; }
+        }
+
         [Fact]
         public void Find_ByRelationId_Success()
         {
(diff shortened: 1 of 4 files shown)


for7raid:
Hit @lbnascimento ! what about this one RP?

Issue #1772, opened 2020-08-02
[BUG]Maximum number of transactions reached


5d2:
V5.08

Exception **Maximum number of transactions reached** 

**CODE**

`var col = _db.GetCollection<T>(typeof(T).Name);
                result = col.Find(bse).ToList();`
`var col = _db.GetCollection<T>(typeof(T).Name);
                t = col.FindById(new ObjectId(Id)) :`

lbnascimento:
@5d2 I could not reproduce the issue. What .NET implementation are you using? Could you provide a complete sample code that I could use to reproduce the issue?

mookid8000:
In case this could help, I can report having experienced the following: Following some fairly heavy load, the following error error occurred:
```
System.ArgumentException: An entry with the same key already exists.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.SortedList`2.Add(TKey key, TValue value)
   at LiteDB.Engine.BasePage.Defrag()
   at LiteDB.Engine.BasePage.InternalInsert(UInt16 bytesLength, Byte& index)
   at LiteDB.Engine.BasePage.Update(Byte index, UInt16 bytesLength)
   at LiteDB.Engine.DataPage.UpdateBlock(DataBlock currentBlock, Int32 bytesLength)
   at LiteDB.Engine.DataService.<>c__DisplayClass4_0.<<Update>g__source|0>d.MoveNext()
   at LiteDB.Engine.BufferWriter..ctor(IEnumerable`1 source)
   at LiteDB.Engine.DataService.Update(CollectionPage col, PageAddress blockAddress, BsonDocument doc)
   at LiteDB.Engine.LiteEngine.UpdateDocument(Snapshot snapshot, CollectionPage col, BsonDocument doc, IndexService indexer, DataService data)
   at LiteDB.Engine.LiteEngine.<>c__DisplayClass26_0.<Update>b__0(TransactionService transaction)
   at LiteDB.Engine.LiteEngine.AutoTransaction[T](Func`2 fn)
   at FleetManager.Model.Tech.LiteDb.LiteDbProjection`1.<Flush>d__16.MoveNext() in C:\projects-rebusfm\fm7-server\FleetManager.Model.Tech\LiteDb\LiteDbProjection.cs:line 159
```
when calling `ILiteCollection<T>.Update(IEnumerable<T> entities)`.

After having thrown that error for a while, LiteDB started giving off error messages like
```
[...]

PR #1906, opened 2020-12-26
Fixed issue #1772: Unfortunately, the TransactionService for a closed…

frankneumann:
… thread was not be removed from the _transactions dictionary within the TransactionMonitor.

Therefore a finalizer for the TransactionService was added, which takes care of removing the transaction from the _transactions dictionary.


mbdavid:
Thanks @frankneumann !

PR #1968, opened 2021-03-25
Fix NullPointer for shell help command

simonhoss:
When you try to run the shell help command you get the exception Object reference not set to an instance of an object.

The reason is that `CustomAttributeExtensions.GetCustomAttributes` where missing the type where to search

Example:
```Welcome to LiteDB Shell

Getting started with `help`

> help
# LiteDB Shell Command Reference

Object reference not set to an instance of an object.
>
```


kcsombrio:
Hi @simonhoss, thanks for this fix

PR #1969, opened 2021-03-27
Offer an option to upgrade streams.

Firedragonweb:
If a litedb is opened from a stream, there is currently no way to upgrade it to V5. I have added an optional upgrade parameter and implemented the upgrade mechanism for writeable streams.

### DIFF
--- a/LiteDB.Tests/Database/Upgrade_Tests.cs
+++ b/LiteDB.Tests/Database/Upgrade_Tests.cs
@@ -88,5 +88,38 @@ namespace LiteDB.Tests.Database
                 }
             }
         }
+
+        [Fact]
+        public void Migrage_From_V4_Stream()
+        {
+            // v5 upgrades only from v4!
+
+            var original = "../../../Utils/Legacy/v4.db";
+
+            using (FileStream fs = new FileStream(original, FileMode.Open, FileAccess.Read))
+            using (MemoryStream ms = new MemoryStream())
+            {
+                fs.CopyTo(ms);
+                ms.Flush();
+                ms.Seek(0, SeekOrigin.Begin);
+
+                using (var db = new LiteDatabase(ms, upgrade:true))
+                {
+                    // convert and open database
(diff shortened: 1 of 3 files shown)


JKamsker (changes requested):
Merge-readiness update for `191c1e789643`.

Changes needed before merge into the v5/master line:

- Preserve the existing public stream-constructor signature for compiled consumers. Adding an optional parameter changes the CLR signature; retain the old overload and forward to a new overload/API.
- Redesign replacement so a failed write cannot destroy the source. The current upgrade truncates the caller's stream before copying the replacement. For arbitrary streams, prefer a separate destination/result unless a recoverable replacement contract can actually be guaranteed; restoring a backup to the same failing stream is not by itself a guarantee.
- Add fault-injection coverage for conversion, copy, flush and stream capability failures, alongside successful encrypted/plain upgrades. Specify stream ownership and position behavior.

The six successful-path upgrade tests passed after conflict resolution, but do not cover constructor binary compatibility or replacement failures. Keep the existing v5 target.

JKamsker:
Addressed the requested stream-upgrade changes in `90042931`:

- restored the exact public `LiteDatabase(Stream, BsonMapper, Stream)` constructor signature;
- replaced in-place truncation with `LiteEngine.Upgrade(source, destination, ...)`, staging conversion before destination replacement and documenting ownership/position/failure behavior;
- supported plain and encrypted v4 stream sources without disposing or modifying the source;
- added coverage for constructor compatibility, plain/encrypted success, current-format no-op, conversion/copy/flush faults, and stream capability validation.

Validation:
- focused upgrade/Issue2494 suite: 14 passed
- `dotnet build LiteDB.sln -c Release -p:TestingEnabled=true`: succeeded
- full Release solution tests: 200 passed, 9 skipped, 0 failed

`tests.runsettings` is not present on the PR base, so the full suite used the checked-in xUnit configuration.

Issue #1982, opened 2021-04-09
Allow custom TypeDescriptor field


skadefro:
**Is your feature request related to a problem? Please describe.**
I want to use lidedb as a local cache for a system that uses mongodb as a backed.
But I already use _type for something else and that gives issues with litedb.

**Describe the solution you'd like**
I would like to suggest adding a TypeDescriptor property on the BsonMapper class to allow a custom name for this field

**Describe alternatives you've considered**
Have not considered any alternatives

**Additional context**
I have opened a pull request with the suggested feature.

mbdavid:
Hi @skadefro,

If you are mapping documents between two databases that need this same `_type` column you need to consider create a custom `BsonMapper` class that use this new rules. You need to use a single object instance of this class in your `LiteDatabase` `ctor`.

skadefro:
Hi mbdavid, I'm sorry if I'm missing something obvious, but I tried copying the BsonMapper class with my updates, to my own project, so I can use it in the LiteDatabase constructor. 
But that will not work since LiteDB.Constants, LiteDB.Reflection, LiteDB.LinqExpressionVisitor and maybe more are all internal.

skadefro:
I created a fork with this work around. I'll close this issue to not keep issue open that will not be fixed/implemented.

PR #1983, opened 2021-04-09
Allow custom TypeDescriptor field #1982

skadefro:
Suggestion for adding custom TypeDescriptor field as mentioned in #1982

### DIFF
--- a/LiteDB/Client/Shared/SharedEngine.cs
+++ b/LiteDB/Client/Shared/SharedEngine.cs
@@ -15,7 +15,7 @@ namespace LiteDB
     public class SharedEngine : ILiteEngine
     {
         private readonly EngineSettings _settings;
-        private readonly Mutex _mutex;
+        private readonly Semaphore _semaphore;
         private LiteEngine _engine;
         private int _stack = 0;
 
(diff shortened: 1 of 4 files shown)


LiteDB/Client/Shared/SharedEngine.cs
```diff
@@ -15,7 +15,7 @@ namespace LiteDB
     public class SharedEngine : ILiteEngine
     {
         private readonly EngineSettings _settings;
-        private readonly Mutex _mutex;
+        private readonly Semaphore _semaphore;
```
JKamsker, line 18: This is not part of the issue. Revert this please.

skadefro: reverted

JKamsker:
Can you also add some tests please? Thanks :)

JKamsker:
Thanks for the PR, and sorry it sat so long. The linked request #1982 was closed in favour of a fork, and this change reads the name from BsonMapper.Global rather than from the mapper instance and has no tests, so we're closing it. A configurable discriminator field fits the mapper-options roadmap in #2831, where we'll consider it together with the other polymorphism work.

PR #1985, opened 2021-04-12
Update BsonMapper.cs

LordOfXen:
Provides attribute flexibility for big projects where wrapping is a thing (e.g. wrapping the LiteDB to a library of the X project with custom attributes).


JKamsker:
Thanks for the PR, and sorry for the very long wait. Entity mapping was restructured on dev and this diff no longer applies. Custom attributes can already be honoured through the BsonMapper.ResolveMember callback, which can rename, ignore or set AutoId on any member. Broader attribute and mapping-option work is collected in #2831.

Issue #1995, opened 2021-04-29
[BUG]


gzatravkin:
**Version**
5.0.9

Which LiteDB version/OS/.NET framework version are you using. **(REQUIRED)**
5.0.9, Windows 10, Turkish settings

**Describe the bug**
Creation of expression with "Min" token causes exception on Turkish systems

**Code to Reproduce**
`            var expression = BsonExpression.Create("{Count: SUM(*.COALESCE(Count, 1)), Citizen: Min(*.Citizen)}"); 
`
**Expected behavior**
Min token should work at turkish systems without errors

**Screenshots/Stacktrace**
```
LiteDB.LiteException: Method 'MİN' does not exist or contains invalid parameters
   at LiteDB.BsonExpressionParser.TryParseMethodCall(Tokenizer tokenizer, ExpressionContext context, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpressionParser.ParseSingleExpression(Tokenizer tokenizer, ExpressionContext context, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpressionParser.ParseFullExpression(Tokenizer tokenizer, ExpressionContext context, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpressionParser.TryParseDocument(Tokenizer tokenizer, ExpressionContext context, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpressionParser.ParseSingleExpression(Tokenizer tokenizer, ExpressionContext context, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpressionParser.ParseFullExpression(Tokenizer tokenizer, ExpressionContext context, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpression.ParseAndCompile(Tokenizer tokenizer, BsonExpressionParserMode mode, BsonDocument parameters, DocumentScope scope)
   at LiteDB.BsonExpression.Create(Tokenizer tokenizer, BsonExpressionParserMode mode, BsonDocument parameters)
   at LiteDB.BsonExpression.Create(String expression, BsonDocument parameters)
   at LiteDB.BsonExpression.Create(String expression)

```

kcsombrio:
Hi @gzatravkin, just commited to master branch the fix to this issue. It will be available in the next LiteDB release.

PR #1996, opened 2021-04-29
replace ToUpper with ToUpperInvariant

gzatravkin:
Fix for https://github.com/mbdavid/LiteDB/issues/1995. It
Replaces ToUpper with ToUpperInvariant, so Min can't be transformed to MİN


kcsombrio:
Hi @gzatravkin, thanks for this PR, but it is currently not passing the unit tests. We are making some changes to fix issue #1995 with ToUpperInvariant() only in the correct spots. Please let us know back in the issue #1995 if it's a fix for your scenario.

Issue #1546, opened 2020-03-09
[QUESTION] Object synchronization method was called from an unsynchronized block of code


jokker23:
I'm getting this error intermittently while running with ConnectionType.Shared and don't with the default. 

```
System.ApplicationException: Object synchronization method was called from an unsynchronized block of code.

   at System.Threading.Mutex.ReleaseMutex()
   at LiteDB.SharedEngine.CloseDatabase()
   at LiteDB.SharedEngine.Upsert(String collection, IEnumerable`1 docs, BsonAutoId autoId)
```
```
System.ApplicationException: Object synchronization method was called from an unsynchronized block of code.
   at System.Threading.Mutex.ReleaseMutex()
   at LiteDB.SharedEngine.CloseDatabase()
   at LiteDB.SharedDataReader.Dispose()
   at LiteDB.LiteQueryable`1.ToDocuments()+MoveNext()
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToList()
```

I tried researching this but haven't come up with any ideas.
Does anyone have any suggestions on why I'm getting this?

JensSchadron:
@jokker23 Could you please provide us with more information like, which version of LiteDB you're using, OS, net runtime, et cetera 🙂

jokker23:
Certainly,

LiteDB: v5.0.2
Win10
.NET Core 3.1

Also using 
Newtonsoft.Json 12.0.3
RabbitMQ.Client 5.1.1
Microsoft.Extensions.Caching.Memory 3.1.1
Microsoft.Extensions.Configuration.Abstractions 3.1.1
Microsoft.Extensions.Configuration.Json 3.1.1
Microsoft.Extensions.DependencyInjection 3.1.1
Microsoft.Extensions.Logging 3.1.1
Microsoft.Extensions.Logging.Console 3.1.1
NLog.Extensions.Logging 1.6.1
And a custom SDK dll

Is there other information that is interesting?

lbnascimento:
@jokker23 Could you provide a sample code that we may use to replicate the issue?

lbnascimento:
@jokker23 Could you test with the current master? The transaction model was entirely rewritten.

jokker23:
Yes sorry I will try to accommodate both requests as soon as possible. I'm currently on overload making VPN tunnels for home workers, sorry for the delay.

alonstar:
I got the same error, too.

[...]

PR #1998, opened 2021-05-02
Added better mutex support for .NET Core (SharedMode)

rexcardan:
Fixed bug where shared mode didn't work with .NET Core - Issue 1707, 1918

### DIFF
--- a/LiteDB.Tests/Engine/Recursion_Tests.cs
+++ b/LiteDB.Tests/Engine/Recursion_Tests.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using Xunit;
+
+namespace LiteDB.Tests.Engine;
+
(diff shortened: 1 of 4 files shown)


haiduong87:
@rexcardan 
Thank you for making this PR
Can you change to net462, because 461 is end of support now
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

Inurias:
Hi, do you think this could also fix https://github.com/mbdavid/LiteDB/issues/1546?

rexcardan:
I think so. I looks like the same symptom.

haiduong87:
@Inurias 
I don't think so, #1546 is another problem

PR #2024, opened 2021-07-07
Refactoring query executor to ensure that transaction is released

gzatravkin:
It removes some duplications and fixes the problem that when you are using 
query.Select(...).FirstOrDefault() won't release the transaction, and eventually, it will lead to a crash when trying to enter exclusive mode. 

It was happening because it wasn't advancing after                             
yield return _query.Select.ExecuteScalar(_pragmas.Collation).AsDocument;

Now releases the transaction inside of finally block, which will be called with Dispose method

### DIFF
--- a/LiteDB/Engine/Query/QueryExecutor.cs
+++ b/LiteDB/Engine/Query/QueryExecutor.cs
@@ -63,75 +63,65 @@ namespace LiteDB.Engine
 
             IEnumerable<BsonDocument> RunQuery()
             {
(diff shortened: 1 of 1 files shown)


JKamsker:
Thanks for tracking this down! On dev the query executor now releases the cursor and transaction from a dispose hook, so stopping after FirstOrDefault() no longer leaks the transaction, and this master-based change no longer applies. If you still see a stuck lock on a v6 prerelease, it's most likely the cross-thread case tracked in #2790.

PR #2087, opened 2021-10-21
fix bug double.NaN

AndrewJ-DEV:
// Creates a simple document
BsonDocument doc = new BsonDocument();
doc["Test"] =  double.NaN ;

// Convert to JSON
string json = JsonSerializer.Serialize(doc);

// Convert back to document, exception "Unexpected token `NaN` in position 9" is throw
BsonDocument doc2 = JsonSerializer.Deserialize(json).AsDocument;

### DIFF
--- a/LiteDB/Document/Json/JsonReader.cs
+++ b/LiteDB/Document/Json/JsonReader.cs
@@ -97,6 +97,8 @@ namespace LiteDB
                         case "null": return BsonValue.Null;
                         case "true": return true;
                         case "false": return false;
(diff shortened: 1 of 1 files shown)


kcsombrio:
Hi @AndrewJ-DEV, as javascript JSON.stringify() converts NaN to 'null', I think the correct implementation would be to do the same. So I'll be closing this pull request and we'll be implementing the conversion from NaN and Infinity to 'null' to keep JSON standards.

Issue #1893, opened 2020-12-09
[BUG] System.UnauthorizedAccessException when running from a shared directory


hche:
**Version**
LiteDB 5.0.9
.NET Framework 4.8

**Describe the bug**
An System.UnauthorizedAccessException is thrown on a write operation to the LiteDB database when the program is started from a shared directory. Running from a local folder everything works as expected. Access rights to the folder are apparently not the problem as the db file is succesfully created.

When the connection mode is set to "shared" the db file is created and the exception is thrown on the 2nd insert operation. On connection mode direct the program creates the db file and inserts the items. When started for the second time, the exception is thrown on leaving the using block (when the items should be written to the db).

**Code to Reproduce**
```
        class Program
        {
            static void Main(string[] args)
            {
                using (var db = new LiteDB.LiteDatabase($"Filename=test.db;Connection=shared"))
                {
                    var collection = db.GetCollection<TestClass>();
                    for (int i = 0; i < 5000; i++)
                    {
                        var test = new TestClass
                        {
                            Id = Guid.NewGuid().ToString(),
                            Name = Path.GetTempFileName()
                        };
                        collection.Insert(test);
                    }
                }
            }
        }
        class TestClass
        {
            public string Id { get; set; }
            public string Name { get; set; }
        }
```

**Expected behavior**
The write operations to the db should not fail on a remote directory. 

**Screenshots/Stacktrace**
```
System.UnauthorizedAccessException: 
   bei System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   bei System.IO.__Error.WinIOError()
   bei System.IO.FileStream.FlushOSBuffer()
   bei System.IO.FileStream.Flush(Boolean flushToDisk)
   bei LiteDB.StreamExtensions.FlushToDisk(Stream stream)
   bei LiteDB.Engine.DiskService.Write(IEnumerable`1 pages, FileOrigin origin)
   bei LiteDB.Engine.WalIndexService.CheckpointInternal()
   bei LiteDB.Engine.WalIndexService.TryCheckpoint()
   bei LiteDB.Engine.LiteEngine.Dispose(Boolean disposing)
   bei LiteDB.SharedEngine.CloseDatabase()
   bei LiteDB.SharedEngine.Insert(String collection, IEnumerable`1 docs, BsonAutoId autoId)
   bei LiteDB.LiteCollection`1.Insert(T entity)
[...]

PR #2096, opened 2021-11-05
[BUG] #1893

gan4-ua:
System.UnauthorizedAccessException when running from a shared directory

I didn't actually manage to find any place when you are flushing the files stream without flushing the OS buffer, so it seems to me like you are simply trying to use the unbuffered I/O.

Just followed the recommendations on MSDN:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers

> Due to disk caching interactions within the system, the FlushFileBuffers function can be inefficient when used after every write to a disk drive device when many writes are being performed separately. If an application is performing multiple writes to disk and also needs to ensure critical data is written to persistent media, the application should use unbuffered I/O instead of frequently calling FlushFileBuffers. To open a file for unbuffered I/O, call the CreateFile function with the FILE_FLAG_NO_BUFFERING and FILE_FLAG_WRITE_THROUGH flags. This prevents the file contents from being cached and flushes the metadata to disk with each write. For more information, see CreateFile.


JKamsker:
Thanks for the PR and the research into FlushFileBuffers. Switching every data file to unbuffered write-through I/O changes durability and alignment requirements on all platforms, and dev is moving towards stronger flushing (#2818), so we won't take this approach. The network-share failure is tracked in #2799, where the plan is a targeted fallback when Flush(true) is rejected.

PR #2126, opened 2021-12-20
Made DiskWriterQueue more reliable so that it can handle multiple back calls better

kaan-kaya:


kcsombrio:
Hi kaan-kaya, thanks for finding and reporting this issue, it's a difficult one to find and simulate. Also thanks for providing a solution to it, that's really good stuff.
About your pull request, I noticed it doesn't have the "_stream.FlushToDisk();" anymore, any special reason for that or just missed it? If it was the latter case, do you want to change your code? Otherwise we could accept it and change it here.
Thanks for using and supporting LiteDB.

kaan-kaya:
@kcsombrio thanks for the review, and thanks for spotting, must have removed that by mistake when cleaning up, I've made an update.

PR #2206, opened 2022-06-24
Changes to properly support Mutexes on NET6.0

conradmicallef:
Global Mutexes were being created without insufficient rights for a 2nd process running on same machine to use them

This fix addresses that by using same methodology applied to NETFRAMEWORK but to NET6.0

### DIFF
--- a/LiteDB/Client/Shared/SharedEngine.cs
+++ b/LiteDB/Client/Shared/SharedEngine.cs
@@ -87,7 +87,7 @@ namespace LiteDB
             }
         }
 
-#region Transaction Operations
+        #region Transaction Operations
 
         public bool BeginTrans()
         {
@@ -139,9 +139,9 @@ namespace LiteDB
             }
         }
 
-#endregion
+        #endregion Transaction Operations
 
-#region Read Operation
+        #region Read Operation
 
         public IBsonDataReader Query(string collection, Query query)
         {
@@ -180,9 +180,9 @@ namespace LiteDB
             }
         }
 
-#endregion
+        #endregion Read Operation
 
-#region Write Operations
+        #region Write Operations
 
         public int Checkpoint()
         {
@@ -352,7 +352,7 @@ namespace LiteDB
             }
         }
 
-#endregion
+        #endregion Write Operations
 
         public void Dispose()
         {
--- a/LiteDB/Client/Shared/SharedEngine.cs
+++ b/LiteDB/Client/Shared/SharedEngine.cs
@@ -5,6 +5,8 @@ using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Threading;
+using LiteDB.Client.Shared;
+
 #if NETFRAMEWORK || NETSTANDARD2_0_OR_GREATER || NET6_0_OR_GREATER
 using System.Security.AccessControl;
 using System.Security.Principal;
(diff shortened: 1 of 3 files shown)


inbarbarkai:
Is there any plan on merging this PR?

tjmoore:
Any update? What's the symptom without this fix, is there an particular error raised?

Looking for solutions to access denied errors on shared mode access, but I'm not seeing it on the mutex but on the journal file (and only one process is writing, the other opens database in read mode).

JKamsker:
I would also like to see some repro code before merging this.

inbarbarkai:
@JKamsker I can't provide a repro, but I can provide the scenario.

Environment: Windows 10/11
Version: 5.0.12

2 writer processes from different users.
1 of the writers is also a reader.

This happened sometimes during our tests, but our customers encountered it all the time, forcing us to release a patch disabling some functionalities.
We replaced it with SQLite to restore those functionalities.

JKamsker:
@inbarbarkai Oh i remember now, i had a similar problem with named pipes and different users.

Ok thank you for the clarification! I will try to get to it asap :)

tjmoore:
Similar to above, have not reproduced the issues we see, but have had them from customers, particularly this kind of error.

`System.IO.IOException: The process cannot access the file 'C:\\XXX\\somedatabase.db' because it is being used by another process.`

This occurs either on .db or the log, and the processes all open the database in shared mode. Have tried long and short lived instances of LiteDatabase also (I believe in shared mode it is internally opening and closing connections on each operation so likely makes little difference, unlike direct mode).

We've done extensive diagnostics and cannot see what the other process is. Some customers show signs of things like Microsoft Defender for Endpoints (the corporate version of Defender) access files but looks more like it detecting file access not actually opening and locking files. Other customers don't have anything like this though.

Any solution needs to be proven fix to roll out to customers though as already got frustrated customers, hence repo is very desirable.

Only alternative have thought about is to move to a single process for database operations and some communication between them. Getting into building a database server territory though and already considering supporting MongoDB as an alternative to LiteDB.

LiteDB/Client/Shared/SharedEngine.cs
```diff
@@ -27,16 +27,33 @@ public SharedEngine(EngineSettings settings)
...
+                ThreadingAclExtensions.SetAccessControl(_mutex, securitySettings);
+#endif
 #else
                 _mutex = new Mutex(false, "Global\\" + name + ".Mutex");
+#endif
+#if NET6_0_OR_GREATER
+                }
 #endif
             }
             catch (NotSupportedException ex)
```
JKamsker, line 59: Big nope! Please make it more readable.

LiteDB/LiteDB.csproj
```diff
@@ -1,18 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.3;netstandard2.0</TargetFrameworks>
-    <AssemblyVersion>5.0.12</AssemblyVersion>
-    <FileVersion>5.0.12</FileVersion>
-    <VersionPrefix>5.0.12</VersionPrefix>
+    <TargetFrameworks>netstandard1.3;netstandard2.0;net6.0</TargetFrameworks>
```
JKamsker, line 4: We are not going to add or drop support for any TargetFrameworks just yet ;)

JKamsker (changes requested):

JKamsker:
I am not sure about the added dependencies. 
about the framework change: we are currently internally thinking about following the official microsoft support matrix.

jobrien9:
Hi all - Is there any update regarding following the Microsoft support matrix? I'd love to add .NET 8 support if that's allowed. I was able to get this working locally for my  .NET 8 applications using [the code from this related issue](https://github.com/mbdavid/LiteDB/issues/2439), but I would be happy to help get this merged if I can.

JKamsker:
Replaced by https://github.com/litedb-org/LiteDB/pull/2711

Issue #2223, opened 2022-08-11
[BUG] Serializer does not use type parameter


marinusbeck:
**Version**
LiteDB version 5.0.12 / Windows 11 / .Net framework 4.7

**Describe the bug**
When serializing a derived class as the base class type it gets serialized as the derived type.

**Code to Reproduce**
```
        class BaseClass
        {
            public int A { get; set; }
        }

        class DerivedClass : BaseClass
        {
            public int B { get; set; }
        }

        [Fact]
        public void Serialize_covariant_type_uses_type_parameter()
        {
            var derived = new DerivedClass()
            {
                A = 1,
                B = 2,
            };

            var mapper = new BsonMapper();

            var docA = mapper.Serialize<BaseClass>(derived);
            var docB = mapper.Deserialize<DerivedClass>(docA);

            Assert.True(docB.B == 0); // expected == 0, actual == 2
        }
```

**Expected behavior**
The expected would be 0 in the above assert, but is actually 2. This is because the mapper reflects on the input object and not the type parameter when serializing.

JKamsker:
This is intended. `BsonMapper` always serializes the object's runtime type. The generic argument only decides whether a `_type` field is written: when the declared type (`BaseClass`) differs from the actual type (`DerivedClass`), the document gets `_type` plus all of `DerivedClass`'s members ([code](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Client/Mapper/BsonMapper.Serialize.cs#L181-L192)). That's what lets polymorphic properties and `List<BaseClass>` collections round-trip to the right subclass. Serializing only the base members would silently drop data there.

If you want a base-only document, project first:

```csharp
var doc = mapper.ToDocument(new BaseClass { A = derived.A });
```

or remove the extra keys (`_type`, `B`) from the resulting `BsonDocument`. You can also register a custom serializer with `mapper.RegisterType<BaseClass>(serialize, deserialize)`, because custom serializers are looked up by the declared type first.

If this doesn't cover your case, comment here and we'll reopen it.

PR #2224, opened 2022-08-11
#2223 changed SerializeObject to use type parameter

marinusbeck:
Fixes serialization of covariant classes based on the type parameter.


JKamsker:
Why would you do that? 
What is the use case here?

What if i am trying to serialize on an Interface and many classes can implement it?
The library currently deserializes as the original type. Even tho you get "IInterface" in reallity the real type stays the same.

Your change removes polymorphism support.

PR #2260, opened 2022-12-07
Change return type of ILiteRepository.Insert to match ILiteDatabase.Insert

InvisiblePhil:
The documentation of ILiteRepository.Insert implied it returns the ID of the insert, but it returned void

Changed to make this API consistent


mahop-net (commented):
Summary should be changed accordingly.

If not accepted summary should also be changed... :-)

InvisiblePhil:
> Summary should be changed accordingly.
> 
> If not accepted summary should also be changed... :-)

The summary as-is notes the method returns the ID - so what change is needed if the PR is accepted?

mahop-net:
OK, it really returns the ID - sorry I thought BsonValue holds the complete document, but it ist only the ID...

Issue #2017, opened 2021-06-14
[BUG] The read lock is being released without being held


laurentiucocanu:
**Version**
5.0.10/Windows 10/NET 5

**Describe the bug**
Trying to read and update records in a parallel foreach throws an exception with the following message: "The read lock is being released without being held."
It seems to run fine in 5.0.3, but not in 5.0.10.

**Code to Reproduce**
        
        public class Target
        {
            [BsonId]
            public ObjectId Id { get; set; }

            public string Name { get; set; }

            public DateTime LastUpdateCheck { get; set; }
        }

        [Fact]
        public void InsertReadTest()
        {
            var dbFullPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
            try
            {
                using (var db = new LiteDatabase(dbFullPath))
                {
                    var table = db.GetCollection<Target>("targets");
                    for (int i = 0; i < 10000; i++)
                    {
                        table.Insert(new Target
                        {
                            Name = $"Name_{i}",
                            LastUpdateCheck = DateTime.UtcNow
                        });
                    }
                    var targetIds = table.FindAll().Select(t => t.Id);
                    Parallel.ForEach(targetIds, targetId =>
                    {
                        var target = table.FindOne(ur => ur.Id == targetId);
                        target.LastUpdateCheck = DateTime.UtcNow;
                        table.Update(target);
                    });
                }
            }
            finally
            {
                File.Delete(dbFullPath);
            }
        }

**Expected behavior**
No exception thrown.

**Screenshots/Stacktrace**
The read lock is being released without being held.

   at System.Threading.Tasks.TaskReplicator.Run[TState](ReplicatableUserAction`1 action, ParallelOptions options, Boolean stopOnFirstFailure)
   at System.Threading.Tasks.Parallel.PartitionerForEachWorker[TSource,TLocal](Partitioner`1 source, ParallelOptions parallelOptions, Action`1 simpleBody, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
--- End of stack trace from previous location ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw(Exception source)
[...]

Issue #1993, opened 2021-04-26
[BUG] LiteDB v5.1 SharedEngine SlimLock issues 


temnava:
**Version**
Which LiteDB version/OS/.NET framework version are you using. v5.1 current branch state (at 2021-04-26)

**Describe the bug**
LiteDB is rising issues with SlimLock.
First one:
ENSURE: current process already locked at the SharedEngine implementation, within the OpenDatabase method.

Second one:
The read lock is being released without being held.
This one occurs at LiteDB.Engine.LockService.ExtiTransaction()
followin the stack trace, it happens when the TransactionService.Finalize() is called upon disposing it within ReleaseTransaction methon in Transaction Monitor.

**Code to Reproduce**
This is the part of a big solution that access the litedb a lot, no simple example to reproduce

**Expected behavior**
No exception upon file write/read.

**Screenshots/Stacktrace**
![image](https://user-images.githubusercontent.com/13450946/116088296-07c20280-a6a2-11eb-8608-04f859716dbf.png)
![image](https://user-images.githubusercontent.com/13450946/116088431-27592b00-a6a2-11eb-88aa-018c0bb9e0ef.png)

**Additional context**
Any ideas how to work around this? 
Would using the LockRecursionPolicy.SupportsRecursion help for the first issue?

JKamsker:
Consolidated into #2790, which combines this report with the related ones and tracks the fix. Closing this as a duplicate — please follow #2790, and comment there if your case is different.

Issue #1637, opened 2020-04-17
[BUG] FindById throws SynchronizationLockException: The read lock is being released without being held.


huyhu:
**Version**
5.0.5

**Describe the bug**
FindById sporadically throws System.Threading.SynchronizationLockException: "The read lock is being released without being held." 

**Code to Reproduce**
Here how I use FindById -it is hard to reproduce the issue though
```
public T FindById<T>(int id)
{
    var data = database.GetCollection<T>(typeof(T).Name).FindById(new BsonValue(id));

    if (Equals(data, default(T)))
    {
        throw new InvalidOperationException($"Cannot find {typeof(T).Name} with id {id}");
    }
    return data;
}
```

**Expected behavior**
No exceptions.

**Screenshots/Stacktrace**
    "ClassName": "System.Threading.SynchronizationLockException",
    "Message": "The read lock is being released without being held.",
    "Data": null,
    "InnerException": null,
    "HelpURL": null,
    "StackTraceString": "   at System.Threading.ReaderWriterLockSlim.ExitReadLock()\r\n   at LiteDB.Engine.TransactionMonitor.ReleaseTransaction(TransactionService transaction)\r\n   at LiteDB.Engine.QueryExecutor.ExecuteQuery(Boolean executionPlan)\r\n   at LiteDB.Engine.LiteEngine.Query(String collection, Query query)\r\n   at LiteDB.LiteQueryable`1.<ToDocuments>d__26.MoveNext()\r\n   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()\r\n   at 

**Additional context**

huyhu:
@mbdavid I have the fix for this issue, but I cannot publish my branch. Could you give me the permissions to do so?

lbnascimento:
@huyhu Can't you make a pull request?

huyhu:
> @huyhu Can't you make a pull request?

remote: Permission to mbdavid/LiteDB.git denied to huyhu.
fatal: unable to access 'https://github.com/mbdavid/LiteDB.git/': The requested URL returned error: 403

lbnascimento:
@huyhu You should create a fork of LiteDB, make your changes there and then submit a pull request.

huyhu:
> @huyhu You should create a fork of LiteDB, make your changes there and then submit a pull request.

@lbnascimento Could you please review the pull request?

mefateah:
Looks like I had the same issue with FindAll method

[...]

Issue #641, opened 2017-07-02
Unable to cast object of type 'LiteDB.EmptyPage' to type 'LiteDB.HeaderPage'


ArefGihuni:
Closed the app and now it's happening.

```
   at LiteDB.PageService.GetPage[T](UInt32 pageID)
   at LiteDB.PageService.NewPage[T](BasePage prevPage)
   at LiteDB.IndexService.AddNode(CollectionIndex index, BsonValue key, Byte level, IndexNode last)
   at LiteDB.LiteEngine.UpdateDocument(CollectionPage col, BsonDocument doc)
   at LiteDB.LiteEngine.<>c__DisplayClass29_0.<Update>b__0(CollectionPage col)
   at LiteDB.LiteEngine.Transaction[T](String collection, Boolean addIfNotExists, Func`2 action)
   at LiteDB.LiteCollection`1.Update(T document)
   at MyAppAsync()
   at System.Threading.Tasks.Task.Execute()
```

mbdavid:
Are you using master branch version or 3.1.1 from Nuget?

ArefGihuni:
Yes, version 3.1.0 from Nuget .. 3.1.1 was too slow so I downgraded .. 

it's a UWP windows 10 app

ArefGihuni:
I started getting InvalidCastException  
Is there any way to fix it?

```
LiteDB.PageService.GetPage<System.__Canon>(Void* InstParam, UInt32 pageID) in at 0:0 
LiteDB.TransactionService.Commit() in at 0:0 
LiteDB.LiteEngine.Commit() in at 0:0 
LiteDB.LiteEngine.Transaction<System.Int32>$catch$0() in at 0:0 
LiteDB.LiteEngine.Insert(String collection, IEnumerable$1<$BsonDocument> docs) in at 0:0 
LiteDB.LiteEngine.Insert(String collection, $BsonDocument doc) in at 0:0 
LiteDB.LiteCollection$1<System.__Canon>.Insert(__Canon document) in at 0:0 
```

mbdavid:
this error occurs in some concurrency scenarios. I'm working on this to fix

lidanger:
I have added the option FileOptions.WriteThrough when opening the data file at 2017.6.13.
The program runs very well utill the same problem appeared 3 days ago again.It's strange, isn't?

The version of LiteDB is 3.1.1.0 with a modify of adding the option FileOptions.WriteThrough.

The first exception is

[...]

PR #2280, opened 2023-01-25
Update LockService.cs to fix "The read lock is being released without being held" error. Fix for #641, 1637, 1993, 2017

mathis1337:
Fix for:
#641, #1637, #1993, #2017

resolving issue of 
"ClassName": "System.Threading.SynchronizationLockException",
"Message": "The read lock is being released without being held.",

Issue is that the ReaderWriterLockSlim can be released or not set if not needed and so the code as it stands will attempt to call ExitReadLock() and this will throw an unhandled exception error. 

I've added and If statement check to ensure there is exactly a ReadLock and also put it in a try catch in the event the Read is released within the small fraction of time the exit is called, so the remaining execution of code can run as needed. 

I suspect that this ReleaseTransaction method that is referenced is multiple places is being called, or its just the lock is released too soon, but this fixes that until a better solution can be implemented. 

I Know other have had this issue, so it would be good to get the PR in or if there is a better place to resolve this let me know and I'll investigate.

### DIFF
--- a/LiteDB/Engine/Services/LockService.cs
+++ b/LiteDB/Engine/Services/LockService.cs
@@ -53,8 +53,16 @@ namespace LiteDB.Engine
         {
             // if current thread are in reserved mode, do not exit transaction (will be exit from ExitExclusive)
             if (_transaction.IsWriteLockHeld) return;
-
-            _transaction.ExitReadLock();
+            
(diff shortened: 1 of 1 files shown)


Pete-PlaytimeSolutions:
Is there any timeframe for when this fix will make it into the main branch?
It's a blocker for us using this library via the NUGET package, currently we are having to use a local fork just to apply this patch

PR #2301, opened 2023-03-21
Bug fix for https://github.com/mbdavid/LiteDB/issues/2298

superahoy:
Adding flag to EntityMapper that a CTor was defined. This will allow BsonMapper.Deserialize to bypass the regular entity deserialization in favor of the custom defined constructor.


JKamsker:
As stated in the bug, we do not need this because you can use ``RegisterType`` https://github.com/mbdavid/LiteDB/commit/4e20fa86040a28c39206d5033924fbc5676d9131

PR #2344, opened 2023-07-18
Fix AesStream Invalid Password exception in corrupted files

anurakt27:
This fix addresses corruption of data/log file which can lead to "Invalid Password" exception [Issue](https://github.com/mbdavid/LiteDB/issues/2343)

Changes:
+ If *bytes[32..63]* is `0`, then write encrypted 1s to *bytes[32..63]*
+ New unit tests to verify:
   + If AesStream can handle a stream of length less than 8192 bytes
   + If AesStream can handle a stream where *bytes[17..8191]* is `0`
+ Removed the logic from AesStream ctor to add padding of 8175 bytes while creating new "hidden" page for encrypted data/log file.

### DIFF
--- a/LiteDB.Tests/Internals/Aes_Tests.cs
+++ b/LiteDB.Tests/Internals/Aes_Tests.cs
@@ -1,5 +1,8 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using System.Linq;
+using System.Reflection;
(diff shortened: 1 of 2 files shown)


mbdavid:
Thanks @anurakt27 !! I will release a new version with this fix today

PR #2384, opened 2023-11-09
Added full record support.

athendrix:
In theory, this should also work for any class that doesn't have an empty constructor, but does have a constructor that matches the properties in names and types.

To use the [BsonId] attribute on a record property, you need to apply it to the *property*, not the constructor value. So you use [property:BsonId] instead.

```cs
public record Customer([property:BsonId] Guid PoorlyNamedId, string Name, string[] Phones, bool IsActive);
```

But if you just name it the way LiteDB expects, you don't even need to do that much.

```cs
public record Customer(Guid CustomerId, string Name, string[] Phones, bool IsActive);
```


athendrix:
My new code passes all the tests, and still works for records.
It just fixes the logic for constructor finding.

athendrix:
This fixes bug #2385 (test included)

athendrix:
@mbdavid What would you say the are the odds this will get merged at some point? Was the master branch the wrong place to submit these kinds of fixes?

For what it's worth, it's only one function swapped out, I even added a test to illustrate the bug being fixed that fails on the master branch but succeeds in my branch, and I'm sure plenty of people would be happy to be able to have records work as well as classes, even if it weren't fixing an actual bug.

Also, feel free to change my coding style, since it's fairly obviously different before merging.

mbdavid:
Hi @athendrix, thanks for your contribution. I'm reading your code now to undersant better your idea. I also will fix your code style, ok?

PR #2389, opened 2023-11-16
Clarify LiteDB.FSharp version support

TeaDrivenDev:
LiteDB.FSharp does not currently work with v5.


mbdavid:
Thanks @TeaDrivenDev

Issue #2146, opened 2022-02-09
[BUG] Index was outside the bounds of the array. / The method or operation is not implemented.


alexSatov:
**Version**
5.0.11 / win10 / net6.0

**Describe the bug**
Method `ILiteCollection<T>.FindOne(Expression<Func<T, bool>> predicate)` throws exceptions with a large number of parallel calls

**Code to Reproduce**
```
class PortedNumber
{
	[BsonId]
	public int Number { get; set; }
}

static ILiteCollection<PortedNumber> PortedNumbers;

[TestCase(1000000)]
void FindNumbers_Parallel(int n)
{
  var numbers = Enumerable.Range(0, n).Select(_ => Random.Next(0, 1000000000)).ToList();
  var results = new TimeSpan[n];
  var tasks = new Task[n];
  
  for (var i = 0; i < n; i++)
  {
	  var j = i;
	  var number = numbers[i];
  
	  tasks[i] = Task.Run(async () =>
	  {
		  var watch = Stopwatch.StartNew();
  
		  try
		  {
			  PortedNumbers.FindOne(pn => pn.Number == number);
		  }
		  catch (Exception e)
		  {
			  Console.WriteLine($"ERROR ({j}): {e.Message}");
		  }
		  finally
		  {
			  watch.Stop();
		  }
  
		  results[j] = watch.Elapsed;
	  });
  }
  
  Task.WaitAll(tasks);
  
  var sum = results.Aggregate((current, next) => current.Add(next));
  
  Console.WriteLine($"Max: {results.Max()}");
  Console.WriteLine($"Min: {results.Min()}");
  Console.WriteLine($"Avg: {sum / n}");
  Console.WriteLine($"Sum: {sum}");
  Console.WriteLine($"First 10:\r\n{string.Join("\r\n", results.Take(10))}");
}
```

**Expected behavior**
No exceptions

**Screenshots/Stacktrace**

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at LiteDB.Engine.IndexNode.GetNextPrev(Byte level, Int32 order)
   at LiteDB.Engine.IndexService.Find(CollectionIndex index, BsonValue value, Boolean sibling, Int32 order)
   at LiteDB.Engine.IndexEquals.Execute(IndexService indexer, CollectionIndex index)+MoveNext()
   at LiteDB.LinqExtensions.<>c__DisplayClass2_0`2.<<DistinctBy>g___|0>d.MoveNext()
[...]

PR #2398, opened 2023-12-14
#2146 Add check for Next/Prev length for IndexNode when searching for index in IndexService.Find

navferty:
If Next/Prev does not contain page addresses with index according to current level, skip this level

### DIFF
--- a/LiteDB/Engine/Services/IndexService.cs
+++ b/LiteDB/Engine/Services/IndexService.cs
@@ -347,7 +347,7 @@ namespace LiteDB.Engine
 
             for (int i = index.MaxLevel - 1; i >= 0; i--)
             {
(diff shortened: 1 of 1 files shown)


alexbereznikov (approved):

alexbereznikov:
Guys just FYI - we live with this fix since year 2023, so far so good.

JKamsker:
Thanks for digging into this, and for the report that it has held up in production. The short Next/Prev arrays come from a page-cache race in which a reader decodes a buffer that is being recycled for another page, so skipping the level hides the symptom and can return wrong results. That race is tracked in #2817 and fixed by the cache rewrite in #2772, and IndexService.Find has since been rewritten on dev, so we're closing this one.

Issue #2307, opened 2023-04-11
[BUG] LiteDB is locking up in the DiskWriterQueue


dgodwin1175:
**Version**
LiteDB:  5.0.16
.NET Framework:  4.6.2

**Describe the bug**
Requests to read or write a collection hang up and get no response when being used by multiple threads.  The only way to recover from this situation is to kill our process that is hosting the database.
E.g. LiteDatabase.GetCollection<Type>(CollectionName)

**Note that this is becoming a critical issue for our application and our customers, so would appreciate a prompt response if at all possible.**

**Steps to Reproduce**
Our application has concurrent threads that are simultaneously reading and updating documents in the database.  Our LiteDatabase object is a singleton, and we use locking on the object to ensure only one thread is updating the database at a time.
After some time, the threads all block waiting on a lock on the database, and the lock is currently held by the above request to GetCollection(), which never returns.

I have downloaded the latest LiteDB code and added some logging to pinpoint where the problem is and it appears to be in the **DiskWriterQueue** where the **ExecuteQueue** function is called by Task.Run(ExecuteQueue), however ExecuteQueue is never actually run, and so the queue is never processed.  See below log snippets (these logs are all mine that I have added):

2023-04-11 11:56:58.054 [17] ERROR Got lock in DiskWriter.Run(), queue size is 30
2023-04-11 11:56:58.054 [17] ERROR Run(): oldValue is 0, _running is 1, task is null False
2023-04-11 11:56:58.054 [17] ERROR **Creating new task to ExecuteQueue, queue size is 30**
2023-04-11 11:56:58.054 [17] ERROR Releasing lock in DiskWriter.Run()
2023-04-11 11:56:58.163 [41] ERROR Got lock in DiskWriter.Run(), queue size is 60
2023-04-11 11:56:58.163 [41] ERROR Run(): oldValue is 1, _running is 1, task is null False
2023-04-11 11:56:58.163 [41] ERROR Releasing lock in DiskWriter.Run()
2023-04-11 11:56:58.195 [9] ERROR Got lock in DiskWriter.Run(), queue size is 62
2023-04-11 11:56:58.195 [9] ERROR Run(): oldValue is 1, _running is 1, task is null False
2023-04-11 11:56:58.195 [9] ERROR Releasing lock in DiskWriter.Run()
2023-04-11 11:56:58.210 [38] ERROR Got lock in DiskWriter.Run(), queue size is 63
[...]

PR #2411, opened 2024-01-25
Simplified DiskWriterQueue with blocking concurrency

ltetak:
It is relatively easy to put the DiskWriterQueue into a state where it does nothing. It is caused by mismatches where the logic does not track properly which _task is the current one. It has many problems:
* Wait() waits for a wrong _task
* _Task is not started at all

e.g.
https://github.com/mbdavid/LiteDB/issues/2307

My repro steps were to run a lot of Inserts and Deletes in parallel (to fill up the disk queue). Then every couple of seconds run _db.Checkpoint() to force full db lock and Wait() invocation.

Fix is to use a much simpler blocking approach (one thread is dedicated to this). It is a good tradeoff IMO for now. It can be later replaced with an awaitable mutex version.
Edit: I added an async version of the semaphore which does not block the thread.

### DIFF
--- a/LiteDB/Engine/Disk/DiskService.cs
+++ b/LiteDB/Engine/Disk/DiskService.cs
@@ -186,8 +186,6 @@ namespace LiteDB.Engine
                 count++;
             }
 
-            _queue.Value.Run();
-
             return count;
         }
 
--- a/LiteDB/Engine/Disk/DiskWriterQueue.cs
+++ b/LiteDB/Engine/Disk/DiskWriterQueue.cs
@@ -21,13 +21,12 @@ namespace LiteDB.Engine
 
         private readonly ConcurrentQueue<PageBuffer> _queue = new ConcurrentQueue<PageBuffer>();
         private readonly object _queueSync = new object();
-        private readonly ManualResetEventSlim _queueHasItems = new ManualResetEventSlim(false);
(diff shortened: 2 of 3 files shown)


mbdavid:
Thanks! This are an old code that must be updated.

jdtkw:
Thanks @ltetak - this indeed resolved our isue (#2307 - I work with @dgodwin1175), but v5.0.18 and v5.0.19 causes us to hit #2435 prior to being able to validate this with an official build. A custom build of #2436 on top of v5.0.19 (that includes #2411) seems to indicate that we can have a stable solution.

ltetak:
hi @jdtkw, transaction (and especially AutoTransaction class) was the next thing I wanted to take a look at. I know about a couple of problems there.
1. AutoTransaction can fail when reverting the transaction - this is bad by itself but it's double-bad because it hides the original exception.
2. Error handling in transactions is wrong causing wrong counts. #2436 may be a fix to it but we need to be sure the DB is in a good state. There are a lot of "ENSURE" errors. My guess is that some transaction does not return the DB to a valid state and it breaks it. 
We run the database in single threaded mode (we serialize every access to the db by locks) so it must be either a problem in the algorithm somewhere or some external exception. I have some evidence that external exceptions make this problem much worse so I would start there - it means if you have an unstable storage medium causing random exceptions it may lead to a corrupted database (which should not happen thanks to the journal approach).

PR #2413, opened 2024-01-27
Fix Rebuild so that it truncates the database file correctly

hizume:
Rebuild operation truncates the database file too short when the file size is greater than 4 GB.
This fix addresses the above issue.


mbdavid:
Thanks @hizume

PR #2415, opened 2024-02-01
chore: throw exception when encounter unpaired surrogate instead of replace with U+FFFD

### LINKED ISSUE #2406: [BUG] LiteDB replaces unpaired surrogate pair to U+FFFD (REPLACEMENT CHARACTER)
**Version**
Which LiteDB version/OS/.NET framework version are you using. **(REQUIRED)**
- litedb 5.0.17
- macOS 14.2.1 23C71 arm64 
- dotnet 8.0

**Describe the bug**

LiteDB replaces unpaired surrogate pair tp U+FFFD silently.

**Code to Reproduce**
Write a small snippet to isolate your bug and could be possible to our team test. **(REQUIRED)**

```
using (var db = new LiteDatabase("test.litedb"))
{
    var collection = db.GetCollection("test");
    var document = new BsonDocument()
    {
        { "UnpairedSurrogate", "\uD800" },
    };
    Console.WriteLine("Before insert: " + (document["UnpairedSurrogate"].AsString == "\uD800")); // Prints True: OK
    var inserted = collection.FindById(collection.Insert(document));
    Console.WriteLine("After insert: " + (inserted["UnpairedSurrogate"].AsString == "\uD800")); // Prints False: bad. should be True
}
```

**Expected behavior**

One of the following is expected
1. Preserve unpaired surrogate pair after serialization
1. Throw exception on `implicit operator BsonValue(string value)` since not supported by Bson
1. Throw exception on `ILiteCollection.Insert` since not supported by LiteDB

**Screenshots/Stacktrace**
If applicable, add screenshots/stacktrace

**Additional context**

Unpaired surrogate pair is valid for windows path name so I think preserving is the best. 
However, unpaired surrogate pair is not valid character in UTF-8 so I think it's reasonable to not support unpaired surrogate.

For preserving unpaired surrogate pair with backwards compatible (and forward compatible if string is valid UTF8) way, you may use [WTF-8](http://simonsapin.github.io/wtf-8/), an extension of UTF-8 used by [rust](https://github.com/rust-lang/rust/blob/master/library/std/src/sys_common/wtf8.rs) and [go](https://github.com/golang/go/commit/974236bda9b9aad87b4b10ec9af2cc01b14e382f) to preserve unpaired surrogate pair in windows paths.

The reason why LiteDB replaces unpaired surrogate pair to U+FFFD silently is LiteDB uses `Encoding.UTF8` which is `new UTF8Encoding(true, false)`. I think LiteDB should use `new UTF8Encoding(false, true)`.

anatawa12:
Fixes #2406

### DIFF
--- a/LiteDB/Document/BsonValue.cs
+++ b/LiteDB/Document/BsonValue.cs
@@ -648,7 +648,7 @@ namespace LiteDB
                 case BsonType.Double: return 8;
                 case BsonType.Decimal: return 16;
 
-                case BsonType.String: return Encoding.UTF8.GetByteCount(this.AsString);
+                case BsonType.String: return StringEncoding.UTF8.GetByteCount(this.AsString);
 
                 case BsonType.Binary: return this.AsBinary.Length;
                 case BsonType.ObjectId: return 12;
@@ -674,7 +674,7 @@ namespace LiteDB
 
             return
                 1 + // element type
-                Encoding.UTF8.GetByteCount(key) + // CString
+                StringEncoding.UTF8.GetByteCount(key) + // CString
                 1 + // CString \0
                 value.GetBytesCount(true) +
                 (variant ? 5 : 0); // bytes.Length + 0x??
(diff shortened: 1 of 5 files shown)


mbdavid:
Hi @anatawa12, thanks for this contribuition.

PR #2423, opened 2024-02-19
Switch to using PackageLicenseExpression

lahma:
By using license expression it will show up correctly on nuget.org and helps license validators/checkers to know what is the exact license type.


mbdavid:
Thanks for that @lahma... this attribute is new for me

PR #2432, opened 2024-02-20
Add possibility to delete backup file when upgrading database

viveleltsi:
It's maybe not so important but because I want to handle the migration of my databases by a home-made app it will be great if I have the option to disabled the backup when upgrading the database.

I made theses changes as simple as possible.

## Summary by CodeRabbit

- **New Features**
  - Added an option to control backup creation during database upgrades and rebuilds.
  - Original database and log files are retained by default and removed only after a successful replacement when backup retention is disabled.
  - Rebuild operations now report their results, including size changes and backup information.

- **Bug Fixes**
  - Improved recovery safety by preserving original files after failed upgrades, replacement failures, partial rebuilds, and validation errors.
  - Backups remain available when rebuild or import errors occur.

### DIFF
--- a/LiteDB/Engine/Engine/Recovery.cs
+++ b/LiteDB/Engine/Engine/Recovery.cs
@@ -13,7 +13,7 @@ namespace LiteDB.Engine
         /// <summary>
         /// Recovery datafile using a rebuild process. Run only on "Open" database
         /// </summary>
-        private void Recovery(Collation collation)
+        private void Recovery(Collation collation, bool createBackup = true)
         {
             // run build service
             var rebuilder = new RebuildService(_settings);
@@ -21,7 +21,8 @@ namespace LiteDB.Engine
             {
                 Collation = collation,
                 Password = _settings.Password,
-                IncludeErrorReport = true
+                IncludeErrorReport = true,
+                CreateBackup = createBackup
             };
 
             // run rebuild process
--- a/LiteDB/Engine/Engine/Upgrade.cs
+++ b/LiteDB/Engine/Engine/Upgrade.cs
@@ -34,7 +34,7 @@ namespace LiteDB.Engine
             }
 
             // run rebuild process
-            this.Recovery(_settings.Collation);
+            this.Recovery(_settings.Collation, _settings.CreateBackupOnUpgrade);
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


JKamsker (changes requested):
Merge-readiness update for `e892066c15ca`.

Changes needed before merge:

- Add focused tests for the new option: the default retains the original backup; opting out deletes the intended backup only after successful upgrade/rebuild; failures during conversion or replacement leave a recoverable original/backup.
- Define how partial rebuilds with reported errors are treated before deleting the original, and cover that policy in a regression test.
- Specify and test the treatment of log-file backups. Avoid leaving the meaning of `CreateBackup` ambiguous between data and log files.

Existing upgrade/rebuild coverage passed during conflict resolution, but it does not exercise these new deletion branches. Please run the new cases and current CI before marking this ready to merge.

JKamsker:
Addressed the requested backup-safety changes in `ea267795`:

- delete both staged data and log backups when backup retention is disabled and replacement completed without reported errors
- retain the original data/log artifacts for partial rebuilds and conversion or replacement failures
- document the cleanup contract for rebuild and upgrade options
- add focused coverage for default retention, successful opt-out rebuild/upgrade, conversion failure, replacement failure/recovery, partial rebuild errors, and log files

Validation:
- `RebuildBackup_Tests`: 6 passed
- rebuild/upgrade focused suite: 23 passed, 1 existing skip
- Release solution build with test hooks: succeeded
- full local net8.0 suite: 870 passed, 7 skipped; net10.0 suite: 870 passed, 7 skipped; ReproRunner: 18 passed
- local .NET Framework targets could not start because Mono is not installed, while the GitHub Windows matrix covers those platform paths
- GitHub Actions run 34784090030: all 45 build, test, cross-process, and repro jobs passed

The prior inline review thread is resolved.

JKamsker (approved):
All requested backup-safety changes are addressed through `cafd2d40`. Cleanup now waits for successful replacement validation, both backup artifacts follow the documented policy, failure and partial-rebuild paths retain recovery data, and the focused plus full CI suites pass.

JKamsker:
Follow-up for `cafd2d40`:

- deferred opt-out cleanup until the rebuilt/upgraded database completes `Open()` validation
- added a regression proving a post-replacement collation validation failure retains the byte-identical original backup
- made v7 detection fill the version buffer across short stream reads

Validation after the follow-up:
- backup safety suite: 7 passed
- rebuild/upgrade suite: 24 passed, 1 existing skip
- Release solution build: succeeded
- net8.0: 871 passed, 7 skipped
- net10.0: 871 passed, 7 skipped
- ReproRunner: 18 passed
- GitHub Actions run 34784750274: all 45 jobs passed

Both review threads are resolved and CodeRabbit marks the latest inline finding addressed.

PR #2433, opened 2024-02-20
Fix issue: Id is null when anon. type is de/-serialized

JKamsker:
Anonymous types get their _id zeroed out when deserializing.

### DIFF
--- a/LiteDB.Tests/Document/Bson_Tests.cs
+++ b/LiteDB.Tests/Document/Bson_Tests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


mbdavid:
Thanks @JKamsker

JKamsker:
Thanks for the Switft merge 🚀 

Was this even the correct approach? 
Is this the way it's done in the other places?
I didnt find it, that's why i am asking.

I am:
- Deserializing "Id"
- If it is null, then check if _id is there and deserialize that

But it could also be:
- If ``par.Name`` == ``Id`` then deserialize ``_id``

Just not completely sure what the convention here is 😅

PR #2436, opened 2024-02-23
Fix #2435 Transactions are not removed in LiteDB 5.0.18

### LINKED ISSUE #2435: [BUG] Transactions are not removed in LiteDB 5.0.18
**Version**
5.0.18

**Describe the bug**
We use a single IDatabase  instance registered as singleton. Our application performs reads and writes from many different threads. After a little while, LiteDB starts throwing exceptions that the max number of transactions is reached. While inspecting TransactionMonitor, we only see an increasing number of transactions.

With 5.0.17 everything works fine and transactions are released properly.

**Code to Reproduce**
```
for (var i = 0; i < 110; i++)
{
    await Task.Run(() =>
    {
        try
        {
            var result = _collection.FindById(key);
        }
        catch (Exception ex)
        {
            // Do something with exception. After 100 iterations, the max number of transactions will be reached
        }
    });
}

```

**Expected behavior**
The code above should run without exception.

hizume:
Fix #2435

### DIFF
--- a/LiteDB/Engine/Query/QueryExecutor.cs
+++ b/LiteDB/Engine/Query/QueryExecutor.cs
@@ -76,103 +76,97 @@ namespace LiteDB.Engine
 
             IEnumerable<BsonDocument> RunQuery()
             {
-                var snapshot = transaction.CreateSnapshot(_query.ForUpdate ? LockMode.Write : LockMode.Read, _collection, false);
-
-                // no collection, no documents
-                if (snapshot.CollectionPage == null && _source == null)
+                try
                 {
-                    // if query use Source (*) need runs with empty data source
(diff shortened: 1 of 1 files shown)


alexanderdibenedetto:
Hi - can we get some movement on this? It's blocking us from updating to LiteDB 5.0.18+ and getting the auto-rebuild functionality we so desperately would like to see.

viceice (approved):
@mbdavid Can you please have a look into this? If you compare withtout whitespace then the changes are small.

https://github.com/mbdavid/LiteDB/pull/2436/files?diff=unified&w=1

gagmax981 (approved):

JKamsker:
The issue has been fixed.

Issue #2421, opened 2024-02-18
[BUG] Memory Leak


kimdiego2098:
**Version**
5.0.1.7

**Describe the bug**

When using ``Query.LT`` or other code for ``BsonExpressionScalarDelegate`` , looking up the source code may reveal that there is a dictionary storing expressions, but this dictionary cache is the root cause of memory leakage. I think it should be changed to memorycache (i.e. adding timeout clearing). As can be seen from the graph, the number of dictionary elements has reached over 4000

I had to change the Queries to use Lambda expressions.But this is inconvenient, and there may still be different keys in the dictionary, leading to memory leaks

![A_{KP1B6LJ3I)V015VER755](https://github.com/mbdavid/LiteDB/assets/82756760/0493cd67-19f3-47f8-8e55-7b655714b6ac)

**Code to Reproduce**

![image](https://github.com/mbdavid/LiteDB/assets/82756760/a2ab3373-237d-4267-a9e3-327fac413d90)

**Expected behavior**

**Screenshots/Stacktrace**
If applicable, add screenshots/stacktrace

**Additional context**

kimdiego2098:
@mbdavid Sorry to bother you, but this is a serious memory leak issue

JKamsker:
I will be looking into it

JKamsker:
Consolidated into #2807, which combines this report with the related ones and tracks the fix. Closing this as a duplicate — please follow #2807, and comment there if your case is different.

PR #2438, opened 2024-02-26
Fixed memory leak due to infinite static caches

Sasha-hi:
Use MemoryCache with fixed SlidingExpiration.
Increasing the required framework version to 4.5.1
Not the best solution, but better than endlessly eating up memory

#2421

### DIFF
--- a/LiteDB/LiteDB.csproj
+++ b/LiteDB/LiteDB.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net4.5;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net4.5.1;netstandard1.3;netstandard2.0</TargetFrameworks>
     <AssemblyVersion>5.0.19</AssemblyVersion>
     <FileVersion>5.0.19</FileVersion>
     <VersionPrefix>5.0.19</VersionPrefix>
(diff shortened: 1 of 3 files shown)


LiteDB/LiteDB.csproj
```diff
@@ -56,15 +56,24 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'net4.5.1'">
     <Reference Include="System" />
     <Reference Include="System.Runtime" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="1.1.2" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
     <PackageReference Include="System.Reflection.TypeExtensions" Version="4.5.1" />
     <PackageReference Include="System.Security.Cryptography.Algorithms" Version="4.3.1" />
+    <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="1.1.2" />
```
dombrovsky, line 68: I am not a maintainer of this lib, but I doubt adding Microsoft.Extensions.Caching.Memory dependency would be approved. Implementing cache expiration logic relying on what is available in BCL may be needed.
Also, since this is a general purpose library, few concerns regarding sliding cache expiration approach:
1. It introduces implicit dependency on static DateTime.Now or similar, which may become a problem for unit testing (not only LiteDB itself, but any other projects that uses LiteDB in tests). Microsoft.Extensions.Caching.Memory has an ISystemClock abstraction. If LiteDB library would rely on system time for caching logic, there should be a way to supply similar abstraction to LiteDB, so implementation could be substituted if needed.
2. The hard decision to have sliding expiration exactly to 1 minute may need to be parametrized.

Sasha-hi: Thank you for your review. 
"dependency would be approved" - Absolutely agree. I'll try to rewrite this when I have some free time.
"which may become a problem for unit testing" - I absolutely agree, but I still have no idea how to fix this without seriously reworking the project architecture. Static methods and static caches are not the best solution. 
"need to be parametrized" - Thanks, done. At the same time, now by default the parameter does not change the existing logic - indefinite cache

JKamsker:
Although I am not 100% sure, I fear this cache doesn't solve the problem. 
The problem is, that lambda.compile generates a delegate to *something* that might be an interpreted expression or real il code that is actually executed. In the ladder case, removing the reference to the delegate would not cause any relief in the memory pressure.

Would you mind providing unit tests and benchmark(+results)?

JKamsker:
I think the problem this pr addresses is real but its really not solved with this pr. I am going to close it for now.
Feel free to reopen it at any time.

Issue #2446, opened 2024-03-12
[BUG] DiskWriterQueue captures synchronization context


alexbereznikov:
**Version**
LiteDB v5.0.19, Windows 11, .NET 6

**Describe the bug**
`DiskWriterQueue` captures synchronization context/task scheduler due to omitting task scheduler parameter in `Task.Factory.StartNew`. This can cause hard to debug side effects when custom synchronization contexts are being used. For contexts that can become invalid, `Post` call caused by state machine inside `DiskWriterQueue.ExecuteQueue` can break application.

**Code to Reproduce**
Unfortunately, I can't reproduce this issue in a new project.

**Expected behavior**
`DiskWriterQueue` should not capture synchronization context for its own operations.

**Screenshots/Stacktrace**
Part of the stack trace that leads to this error:
```
   at Zero.Foundation.Threading.TaskQueueSynchronizationContext.Post(SendOrPostCallback d, Object state) in C:\s\apollo.desktop\src\Zero.Foundation\Threading\TaskQueueSynchronizationContext.cs:line 40&#xD;
   at System.Threading.Tasks.SynchronizationContextAwaitTaskContinuation.PostAction(Object state)&#xD;
   at System.Threading.Tasks.AwaitTaskContinuation.RunCallback(ContextCallback callback, Object state, Task&amp; currentTask)&#xD;
--- End of stack trace from previous location ---&#xD;
   at System.Threading.Tasks.Task.&lt;&gt;c.&lt;ThrowAsync&gt;b__128_1(Object state)&#xD;
   at System.Threading.QueueUserWorkItemCallback.&lt;&gt;c.&lt;.cctor&gt;b__6_0(QueueUserWorkItemCallback quwi)&#xD;
   at System.Threading.ExecutionContext.RunForThreadPoolUnsafe[TState](ExecutionContext executionContext, Action`1 callback, TState&amp; state)&#xD;
   at System.Threading.QueueUserWorkItemCallback.Execute()&#xD;
   at System.Threading.ThreadPoolWorkQueue.Dispatch()&#xD;
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()&#xD;
   at System.Threading.Thread.StartCallback()&#xD;
```
Logs from custom synchronization context that shows what is being called:
```
POST &lt;.cctor&gt;b__8_0 System.Threading.Tasks.SynchronizationContextAwaitTaskContinuation+&lt;&gt;c&#xD;
[...]

PR #2447, opened 2024-03-12
Don't capture synchronization context in DiskWriterQueue

alexbereznikov:
This addresses #2446 

Task.Factory.StartNew tries to use task scheduler from the current task, so TaskScheduler.Default needs to be passed here.

### DIFF
--- a/LiteDB/Engine/Disk/DiskWriterQueue.cs
+++ b/LiteDB/Engine/Disk/DiskWriterQueue.cs
@@ -50,15 +50,15 @@ namespace LiteDB.Engine
             // throw last exception that stop running queue
             if (_exception != null) throw _exception;
 
+            _queueIsEmpty.Reset();
+            _queue.Enqueue(page);
+            _queueHasItems.Set();
+
             lock (_queueSync)
             {
-                _queueIsEmpty.Reset();
-                _queue.Enqueue(page);
-                _queueHasItems.Set();
-
                 if (_task == null)
                 {
-                    _task = Task.Factory.StartNew(ExecuteQueue, CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default);
+                    _task = Task.Factory.StartNew(ExecuteQueue, TaskCreationOptions.LongRunning);
                 }
             }
         }
@@ -76,7 +76,7 @@ namespace LiteDB.Engine
         /// <summary>
         /// Execute all items in queue sync
         /// </summary>
-        private async Task ExecuteQueue()
+        private void ExecuteQueue()
         {
             try
             {
@@ -88,19 +88,17 @@ namespace LiteDB.Engine
                     }
                     else
                     {
-                        lock (_queueSync)
-                        {
-                            if (_queue.Count > 0) continue;
 
-                            _queueIsEmpty.Set();
-                            _queueHasItems.Reset();
+                        if (_queue.Count > 0) continue;
 
-                            if (_shouldClose) return;
-                        }
+                        _queueIsEmpty.Set();
+                        _queueHasItems.Reset();
+
+                        if (_shouldClose) return;
 
                         _stream.FlushToDisk();
 
-                        await _queueHasItems.WaitAsync().ConfigureAwait(false);
+                        _queueHasItems.WaitAsync().GetAwaiter().GetResult();
                     }
                 }
             }
@@ -137,7 +135,7 @@ namespace LiteDB.Engine
             _shouldClose = true;
             _queueHasItems.Set(); // unblock the running loop in case there are no items
 
-            _task?.Wait();
+            _task?.GetAwaiter().GetResult();
             _task = null;
         }
     }
--- a/LiteDB/Engine/Disk/DiskWriterQueue.cs
+++ b/LiteDB/Engine/Disk/DiskWriterQueue.cs
@@ -58,7 +58,7 @@ namespace LiteDB.Engine
 
                 if (_task == null)
                 {
-                    _task = Task.Factory.StartNew(ExecuteQueue, TaskCreationOptions.LongRunning);
+                    _task = Task.Factory.StartNew(ExecuteQueue, CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default);
                 }
             }
         }
@@ -100,7 +100,7 @@ namespace LiteDB.Engine
 
                         _stream.FlushToDisk();
 
-                        await _queueHasItems.WaitAsync();
+                        await _queueHasItems.WaitAsync().ConfigureAwait(false);
                     }
                 }
             }


pictos (changes requested):
I can see the value of not capture the syncronization context, but this will be a breaking change. From the linked issue, it's caused because your project uses a custom `SyncronziationContext`, but that doesn't happen on UI frameworks, that implements its own syncContext, so I that doesn't prove that the issue is in your SyncContext implementation and not on LiteDb?

alexbereznikov:
@pictos As I described in related issue (#2446), breaking change was introduced after 5.0.17 - Task.Run (which uses default task scheduler) has been changed to Task.Factory.StartNew (that uses a current task scheduler by default). More precisely, breaking changes were made in this commit - https://github.com/mbdavid/LiteDB/commit/f21cd8405eda7c7b59b018c9a541b0c3605eb5da

So this PR aims to revert this breaking change, not to introduce one.

Regardless of whether our custom context is implemented correctly, I don't think it is a good idea to execute writer queue on a captured context (e g UI context). I would say, it rather highlights the issue.

pictos:
So, looking more closely to the implementation I strongly believe that the issue is more than having it capturing the current task's `TaskScheduler`.
The `task` is being created with `LongRunning` flag, and uses an `async delegate`, and you may know that those two doesn't fit togheter.

So the first thing we should change here is to make the `ExecuteQueue` a `void` method, and block the created thread on the async operations or rethink all this engine, which I believe will not be done on this PR if we choose this path.

pictos (commented):
After talking with @mbdavid, I could understand the idea behind this member. So the fix is something completely different from your implementation.

No matter if you pass the TaskScheduler as parameter or not, it will not run on the UI Thread. The fix is making the ExecuteQueue method a void method and all the implementation should be a block code, that's not use any asynchronous API.

Also, for improvement, remove the lock object on the task instantiation and use an interlockedExchange method to verify if there's a need to create a task.

With that the code will behave as expected, it will run on a deticated thread and will not touch any kind of syncContext, taskScheduler or threadPool Thread.

Let me know if you want to perform these changes, otherwise I'll do it. And thanks for your hard work so far❣️

alexbereznikov:
@pictos I saw this discussion (until I was removed from #moderators channel :)). Yes this could solve the issue as well, but what are real benefits of having a separate thread for this? I mean, this is not a like windows message loop thread, where we have to keep the same thread id. From my point of view, having a separate thread here have some drawbacks:
* Having to manage this thread 
* Additional resources are required - at least 1Mb for the thread stack
* Loosing ability to use async primitives - roughly half of the time the thread will do nothing while waiting for event

I agree that LongRunning probably is redundant here, we can just as well revert to Task.Run.

Regarding PoC - I couldn't reproduce the issue in a small PoC either, but I'm 100% sure that our integration tests started to fail because of this regression, and this fix fixed this for us (we have a fork with some fixes applied).

pictos:
@alexbereznikov, let me answer your comments in pieces.

>  Yes this could solve the issue as well, but what are real benefits of having a separate thread for this? I mean, this is not a like windows message loop thread, where we have to keep the same thread id


First of all, we don't need to manage the dedicated thread, that BCL will do it for us, since we're using the `Task.Factory.StartNew`. So, we don't need to manage the thread, we are handling with the `Task`, which is a high-level abstraction.

> Loosing ability to use async primitives - roughly half of the time the thread will do nothing while waiting for event

We don't need the async primitives in this specific case. We want to have a dedicated thread to handle the `DiskWriter` work.

> I agree that LongRunning probably is redundant here, we can just as well revert to Task.Run.

I didn't say the `LongRunning` is redundant, I said that passing the `TaskScheduler` as parameter is redundant. And we **can't** use `Task.Run` here, because that way we will block a `ThreadPool` thread to do a hard work, and that's wrong by design. The threads on the `ThreadPool` aren't designed to do a heavy work like that. That said, the usage of the `Task.Factory.StartNew` is a good approach here. 

> Regarding PoC - I couldn't reproduce the issue in a small PoC either, but I'm 100% sure that our integration tests started to fail because of this regression, and this fix fixed this for us (we have a fork with some fixes applied).

Friendly saying. And still there you're confident that the custom SyncronizationContext implemented on your tests doesn't have an issue?

alexbereznikov:
@pictos Ok, we do use sync I/O operations here and thread pool is really not the best choice in this case.

But, what if we change `WritePageToStream` to use `WriteAsync` instead of `Write`, and `FlushToDisk` extension to use `FlushAsync` instead of `Flush`? In this case, we won't block thread pool thread to do a hard work - its responsibility will be pass pages from queue to _stream.WriteAsync. Also, thread won't wait on `WaitAsync`

pictos:
@alexbereznikov, why are so against to use a dedicated thread for it? It's the right tool this job.

For this PR at this time we have plan to not change how it works in core, so I'll follow the Mauricio's recommendations. And feel free to open an issue proposing this change, if it's approved we can merge it in the next major release.

alexbereznikov:
@pictos I just feel that it is a step in a backwards direction. I don't think that changing Write to WriteAsync can break anything.

Regarding custom sync context and PoC - please take a look at this - https://github.com/alexbereznikov/LiteDB/tree/2446-poc

Here I added a new console project called Poc, which uses a custom synchronization context and a task scheduler created from it. It outputs all the stack traces for each call to Send/Post, and it is clear that ExecuteQueue calls got redirected to this context.

Context itself does not flow, for sure

pictos:
@alexbereznikov I've a feeling that you aren't listening to me and just want to make your point pass. What you did on your PoC just confirms what I said [here](https://github.com/mbdavid/LiteDB/pull/2447#issuecomment-2153079695), what in resume is "the only way that your custom SyncContext will flow to the custom Task is if you call `TaskScheduler.FromSyncronizationContext()`. Passing TaskScheduler.Default or omitting the call will not touch on the SyncContext, no matter what's".

As peer discussion with @mbdavid, for now we will not change the approach. But feel free to open an issue where you can put more info and we can discuss another approach for this, but remember it must be ThreadSafe.

alexbereznikov:
@pictos But in this case Task.Factory.StartNew uses a current task scheduler, which in turn uses that context. I don't think it is a correct behavior, because in this PoC I didn't change anything in LiteDB sources, i e haven't passed this TaskScheduler.FromSyncronizationContext() when starting ExecuteQueue. And passing TaskScheduler.Default fixes this issue.

alexbereznikov:
I'm honestly surprised a lot that such a simple one-line regression fix spawned that much discussion. If you insist I'll do in the way you proposed because I have no other options :) But I'm still 100% sure that this is a step in a backwards direction - instead of fixing the issue and moving on we're rolling back to ways it was done a while ago.

pictos:
@alexbereznikov, let's try one more time... I did a test using winui, which has a syncronizationContext, tried with `master` branch and your code and my changes... The issue right now, that I can see is the `ExecuteQuery` is leaking to the ThreadPool thread. And it shouldn't, right this is the thing that is needed to be fixed and released on next SR.

From the PoC that you shared, the code in question goes to the SyncContext because [YOUR code forces the Task to run on SyncronizationContext](https://github.com/alexbereznikov/LiteDB/blob/2446-poc/Poc/Program.cs#L62), which will influence where the Litedb's code runs. And changing that right now is a breaking change! We can't say how many devs rely on that today and because of that we can't do this change in a SR. So, one more time, please open an issue with this proposal that we way we can merge and schedule it to be in a major release.

Now, we do use a lot of `ManualResetEventSlim` and `lock` code there... So using `TaskScheduler.FromSyncronizationContext()`, at least of a UI software, is the same as asking for a deadlock.

> I'm honestly surprised a lot that such a simple one-line regression fix spawned that much discussion.

Me too, looks like we aren't talking the same English here! You trying to force a change that I already said will not happen on this PR.

pictos:
@alexbereznikov, let's try one more time... Today that API is sensitive to the SyncContext, so can be devs that rely on that. Changing this behavior is a breaking change that we can't do in a SR.

So, for this PR, that will be released in a SR we need to fix the fact that right now the task is leaking to the ThreadPool. For changing what you want, we will need to do a new PR for that and also verify on the code place other places that are sensitive to SyncContexts and change that, hope that now you understand.

Thanks for your help and bringing this to light. I did the necessary changes for now and will create an issue to follow up the next changes.

pictos (approved):

LiteDB/Engine/Disk/DiskWriterQueue.cs
```diff
@@ -88,19 +88,17 @@ private async Task ExecuteQueue()
...
+                        _queueHasItems.Reset();
 
-                            if (_shouldClose) return;
-                        }
+                        if (_shouldClose) return;
 
                         _stream.FlushToDisk();
 
-                        await _queueHasItems.WaitAsync();
+                        _queueHasItems.WaitAsync().GetAwaiter().GetResult();
```
JKamsker, line 101: Why not a normal ManualResetEvent then?

JKamsker (changes requested):

LiteDB/Engine/Disk/DiskWriterQueue.cs
pictos: Probably, but I don't want to change it too much. Substantial changes will happen on #2505

mbdavid (approved):

PR #2453, opened 2024-03-20
Fixed writing pages in the same thread that does EnqueuePage

### LINKED ISSUE #2451: [BUG] Performance degradation in v5.0.18
**Version**
v5.0.18, v5.0.19

**Describe the bug**
Upsert performance degraded in ~14 times in v5.0.18 when compared to 5.0.17

**Code to Reproduce**
Running this snippet with v5.0.17 and 5.0.18 is enough to see the issue. On my machine, 5.0.17 runs in 12 seconds, and 5.0.18 runs in 2m 50 seconds.

```csharp
internal class Program
{
    private const string DbName = "test.db";
    private const int Records = 100_0000;
    private const int Chunk = 1000;

    static void Main(string[] args)
    {
        if (File.Exists(DbName))
        {
            File.Delete(DbName);
        }

        using (var db = new LiteDatabase(
            new ConnectionString
            {
                Filename = DbName,
            }))
        {
            var collection = db.GetCollection<TestRecord>();
            collection.EnsureIndex(x => x.Index);

            var data = GenerateData();

            var sw = Stopwatch.StartNew();

            foreach (var chunk in data)
            {
                collection.Upsert(chunk);
            }

            Console.Write(sw.Elapsed);
        }
    }

    private static IReadOnlyCollection<IReadOnlyCollection<TestRecord>> GenerateData()
    {
        var records = new List<TestRecord>(Records);

        for (var i = 0; i < Records; i++)
        {
            records.Add(new TestRecord
            {
                Index = i,
                First = Random.Shared.Next(),
                Second = Random.Shared.Next(),
            });
        }

        return records.Chunk(Chunk).ToArray();
    }

    private sealed class TestRecord
    {
        public int Index { get; init; }

        public int First { get; init; }

        public int Second { get; init; }
    }
}
```

**Expected behavior**
Expected behavior is that performance in v5.0.18 is more or less the same as in v5.0.17

**Screenshots/Stacktrace**
N/A

**Additional context**
I haven't experimented much, upserting in chunks because that's what we do in our app, maybe chunk size matters, maybe not. Also haven't tried other operations.

alexbereznikov:
AsyncManualResetEvent was using TaskCompletionSource which by default runs continuations synchronously. That led to writing pages in the same thread that called EnqueuePage and, as a result, ~4x performance degradation. 

Used more correct way to asynchronously wait for wait handle.

This partially fixes #2451


JKamsker:
I am concerned about the extra dependency in ``System.Threading.ThreadPool``

alexbereznikov:
Hi @JKamsker , this dependency exists only because of netstandard1.3 target (for some reason thread pool routines are not available out of box in netstandard1.3). I don't think it is a vulnerable package or something.

Benchmark code is contained in this bug - #2451 , as well as results in terms of time taken to run the app. This pr aimed to partially fix it as well

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    internal static class WaitHandleExtensions
+    {
+        public static async Task<bool> WaitAsync(this WaitHandle handle)
+        {
+            var tcs = new TaskCompletionSource<bool>();
```
pictos, line 11:
```suggestion
            var tcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+            using (new ThreadPoolRegistration(handle, tcs))
+            {
+                return await tcs.Task.ConfigureAwait(false);
+            }
+        }
+
+        private sealed class ThreadPoolRegistration : IDisposable
```
pictos, line 19: If we change this to be a `struct` does the code work as expected? If so, let's change it and implement the interface implicit, that way we don't need to box it.

Changing this to a `struct` will avoid allocations if the task is already completed

```suggestion
        private readonly struct ThreadPoolRegistration : IDisposable
```

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            private readonly RegisteredWaitHandle _registeredWaitHandle;
+
+            public ThreadPoolRegistration(WaitHandle handle, TaskCompletionSource<bool> tcs)
+            {
+                _registeredWaitHandle = ThreadPool.RegisterWaitForSingleObject(handle,
+                    (state, timedOut) => ((TaskCompletionSource<bool>)state).TrySetResult(!timedOut), tcs,
+                    Timeout.InfiniteTimeSpan, executeOnlyOnce: true);
```
pictos, line 27: do we want the Timeout to be infinite? I suggest using a large time, for example 1 day. You would be surprise for systems that get deadlocked because of infinite time spans

pictos (changes requested):
Hey @alexbereznikov, thanks for this fix. I've a couple of suggestions.

And please, can you add unit tests for this, that way we can avoid regressions

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
alexbereznikov: Fixed, but this option is not available in .net45

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
alexbereznikov: I can't agree here. What if there is an app that for some reason does not write anything for 2 days? I don't think we can pick any reasonable timeout here. 

Regarding deadlocks - we explicitly call Set() in DiskWriterQueue.Dispose() method, so even with infinite timeout we won't have deadlocks here.

alexbereznikov:
> Hey @alexbereznikov, thanks for this fix. I've a couple of suggestions.
> 
> And please, can you add unit tests for this, that way we can avoid regressions

Hey @pictos , regarding unit test for this - I can't think of a good way to create a unit test for this without exposing DiskWriterQueue internals or coming up with additional abstractions that required only for tests.

I added an ENSURE check instead, so in case of regression - tests will fail.

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
JKamsker: Then use Task.Run for .net45

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+            }
+
+            void IDisposable.Dispose() => _registeredWaitHandle.Unregister(null);
```
pictos, line 33: to avoid boxing we can't implement this explicity

```suggestion
            public void Dispose() => _registeredWaitHandle.Unregister(null);
```

pictos (approved):
LGTM

LiteDB/Utils/Extensions/WaitHandleExtensions.cs
pictos: I think we can leave the net framework as it's

alexbereznikov:
Closed as redundant due to #2515

PR #2454, opened 2024-03-20
Lazy string interpolation in ENSURE

### LINKED ISSUE #2451: [BUG] Performance degradation in v5.0.18

alexbereznikov:
String interpolation on hot paths leads to performance degradation. We need these messages only when error happened.
With #2453 applied this gives 3x performance boost.

This partially fixes #2451

### DIFF
--- a/LiteDB/Engine/Disk/DiskService.cs
+++ b/LiteDB/Engine/Disk/DiskService.cs
@@ -1,10 +1,7 @@
 ﻿using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
-using System.Text;
 using System.Threading;
-using System.Threading.Tasks;
 using static LiteDB.Constants;
 
 namespace LiteDB.Engine
@@ -261,7 +258,7 @@ namespace LiteDB.Engine
 
                     var bytesRead = stream.Read(buffer, 0, PAGE_SIZE);
 
-                    ENSURE(bytesRead == PAGE_SIZE, () => $"ReadFull must read PAGE_SIZE bytes [{bytesRead}]");
+                    ENSURE(bytesRead == PAGE_SIZE, "ReadFull must read PAGE_SIZE bytes [{0}]", bytesRead);
 
                     yield return new PageBuffer(buffer, 0, 0)
                     {
--- a/LiteDB/Engine/Disk/DiskService.cs
+++ b/LiteDB/Engine/Disk/DiskService.cs
@@ -29,7 +29,7 @@ namespace LiteDB.Engine
         private long _logLength;
 
         public DiskService(
-            EngineSettings settings, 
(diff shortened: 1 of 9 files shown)


ltetak:
Good catch. I guess it would be more standard to move the string interpolation deeper as other libraries do it.
public static void ENSURE(bool conditional, string message, params object[] args)

alexbereznikov:
@ltetak Sure I thought about that, but profiling says that most of the time was lost in `BasePage.ToString()` and I'm not quite sure how to avoid calculating it when passing via args

alexbereznikov:
@ltetak Actually this `.ToString()` call is redundant as formatter will call it either way. So it should work in the way you proposed, pushed fixed code, please take a look

ltetak:
@alexbereznikov I am not an owner of this repo. But this looks like a great performance improvement without any risks or side effects.
cc: @mbdavid

JKamsker:
LGTM

Issue #2458, opened 2024-03-23
[BUG] LiteFileStream SetReadStreamPosition causes incorrect seeking


CSDSP:
**Version**
Current master (commit 4f77be4cc096653472bf06ebcf95eb399b1a4ff5).

**Describe the bug**
LiteFileStream.SetReadStreamPosition, part of how seeking is implemented, has a few bugs.
First of all, 
``` 
if (newPosition < 0 && newPosition > Length)
{
    throw new ArgumentOutOfRangeException();
}
```
This is clearly wrong, probably meant to be an or instead of an and. The effect is you can seek to positions below 0 or greater than length, and you get exceptions other than ArgumentOutOfRangeException.

Secondly, and more importantly for my purposes, is the block below:
```
// calculate new chunk position
 _currentChunkIndex = (int)_streamPosition / MAX_CHUNK_SIZE;
_positionInChunk = (int)_streamPosition % MAX_CHUNK_SIZE;
```

This assumes that all chunks aside from the last one are exactly MAX_CHUNK_SIZE bytes. However, that is not the case.
If it is meant to be the case, then the bug is with writing, because various files in my database (uploaded with LiteFileStorage.Upload) have some randomly much smaller chunks.

CSDSP:
On the write size, MemoryStream.Read isn't guaranteed to return the maximum available number of bytes. Currently, a chunk is written for every buffer read, so whenever it happens to read less then max chunk size bytes, it still writes a chunk. This could be fixed, though it would leave the read issue in place for all current databases.

SpaceCheetah:
@alexbereznikov It shows up when you try to seek in a file read stream. Just downloading directly to a file or copying to a memory stream works fine, but if you try to seek there's a chance depending on the file for it to seek to the wrong position. I noticed it when I was getting exceptions where it was trying to read more data from a chunk than actually existed.

Seeking to the beginning still works, seeking anywhere else in the file can fail if any chunks are smaller than max length, which for me is fairly common.

bnuzhouwei:
The Seek get error, so the file can not be suport enableRangeProcessing in MVC file result. How to fix it.!

bnuzhouwei:
It also cause that The LiteFileStream can't be use for read documents with stream by ITextSharp or other document libs.
Hope this fix as soon as possible

JKamsker:
Hello @CSDSP, 
you say there are a few bugs, can you please provide a repro unit test for the bugs you found?

[...]

PR #2459, opened 2024-03-23
Fix LiteFileStream.SetReadStreamPosition

SpaceCheetah:
Fix for issue #2458. Not a great solution, since it makes seeking much more expensive, but it works in my situation. I don't think there's a way to make it significantly faster without making chunks uniform length, which would require changes on the write side and not work for existing databases.

### DIFF
--- a/LiteDB/Client/Storage/LiteFileStream.Read.cs
+++ b/LiteDB/Client/Storage/LiteFileStream.Read.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using static LiteDB.Constants;
@@ -7,6 +8,7 @@ namespace LiteDB
 {
     public partial class LiteFileStream<TFileId> : Stream
     {
+        private Dictionary<int, long> _chunkLengths = new Dictionary<int, long>();
         public override int Read(byte[] buffer, int offset, int count)
         {
             if (_mode != FileAccess.Read) throw new NotSupportedException();
@@ -85,6 +87,7 @@ namespace LiteDB
                 }
                 newChunkIndex++;
             }
+            
             newChunkIndex--;
             seekStreamPosition -= _chunkLengths[newChunkIndex];
             _positionInChunk = (int)(_streamPosition - seekStreamPosition);
--- a/LiteDB/Client/Storage/LiteFileStream.Read.cs
+++ b/LiteDB/Client/Storage/LiteFileStream.Read.cs
@@ -62,11 +62,14 @@ namespace LiteDB
             int loadedChunk = _currentChunkIndex;
             int newChunkIndex = 0;
             while (seekStreamPosition <= _streamPosition) {
-                if (!_chunkLengths.ContainsKey(newChunkIndex)) {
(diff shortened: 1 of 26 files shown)


bnuzhouwei:
What version can use this fix?

LiteDB/Client/Storage/LiteFileStream.Read.cs
```diff
@@ -42,23 +42,40 @@ private byte[] GetChunkData(int index)
...
-            _currentChunkIndex = (int)_streamPosition / MAX_CHUNK_SIZE;
-            _positionInChunk = (int)_streamPosition % MAX_CHUNK_SIZE;
-
-            // get current chunk
-            _currentChunkData = this.GetChunkData(_currentChunkIndex);
+            long seekStreamPosition = 0;
+            int loadedChunk = _currentChunkIndex;
+            int newChunkIndex = 0;
+            while (seekStreamPosition <= _streamPosition) {
+                if (!_chunkLengths.ContainsKey(newChunkIndex)) {
```
JKamsker, line 65: Missing an opportunity of using trygetvalue ;)

JKamsker:
I would like to see some unitttests for this pr as i currently don't fully understand the issue.

LiteDB/Client/Storage/LiteFileStream.Read.cs
```diff
@@ -10,6 +10,9 @@ public partial class LiteFileStream<TFileId> : Stream
         public override int Read(byte[] buffer, int offset, int count)
         {
             if (_mode != FileAccess.Read) throw new NotSupportedException();
+            if (_streamPosition == Length) {
```
JKamsker, line 13: Please use separate lines for the ``{`` and ``}``

LiteDB/Client/Storage/LiteFileStream.cs
```diff
@@ -23,6 +24,7 @@ public partial class LiteFileStream<TFileId> : Stream
         private byte[] _currentChunkData = null;
         private int _positionInChunk = 0;
         private MemoryStream _buffer;
+        private Dictionary<int, long> _chunkLengths = new Dictionary<int, long>();
```
JKamsker, line 27: Not trying to force you change this, but are you sure sure separating fields from their main use is a good idea? I mean if it would be used by many different partials, this would imo the best place but you are only using it in the ``.Read`` partial. 

What do you think? Does that make sense?

JKamsker (approved):

LiteDB/Client/Storage/LiteFileStream.Read.cs
```diff
@@ -42,23 +46,53 @@ private byte[] GetChunkData(int index)
...
+                    seekStreamPosition += length;
+                }
+                else
+                {
+                    loadedChunk = newChunkIndex;
+                    _currentChunkData = GetChunkData(newChunkIndex);
+                    seekStreamPosition += _currentChunkData.Length;
+                }
+                newChunkIndex++;
+            }
```
JKamsker, line 87: While you are at it, add a new line here

JKamsker:
Thank you for your contribution and bearing with me :D

PR #2470, opened 2024-04-10
6 1

ltetak:
Hello the LiteDB community. I would like to open a discussion about the stability of this tool. I have some thoughts first and then there is a draft here about some recent changes + some quick wins.
* Everything seems fairly stable if there are no external issues (usually IOExceptions or corrupted DB)
* Exceptions are not handled very well. Usually not failing fast causing corrupted DB.
* Exceptions are hidden by other exceptions (e.g. AutoTransaction tries to rollback and recover from the exception but if this fails then the original exception is lost)
* There is no easy way from outside to know if the LiteEngine has failed or not - and recover if possible.


LiteDB/Utils/Logging.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Diagnostics;
+using System.Text;
+
+namespace LiteDB
+{
+    public static class Logging
+    {
+        public static event Action<LogEventArgs> LogCallback;
```
ltetak, line 9: Have at least some simple mechanism to hook to logs and log them - to reveal what is wrong if it fails in production

LiteDB/Engine/LiteEngine.cs
```diff
@@ -81,9 +79,14 @@ public LiteEngine(EngineSettings settings)
 
         #region Open & Close
 
+        public bool IsDisposed => _state.Disposed;
+
+        private readonly TaskCompletionSource<bool> _closedTask = new TaskCompletionSource<bool>();
+        public Task<bool> Closed => _closedTask.Task;
```
ltetak, line 85: give a chance to outside observer to check 
if (db.IsDisposed)
{
  await db.Closed;
}
recreate db

LiteDB/Engine/Disk/DiskWriterQueue.cs
```diff
@@ -100,13 +102,13 @@ private async Task ExecuteQueue()
...
 
-                        await _queueHasItems.WaitAsync();
+                        await _queueHasItems.WaitHandle.WaitAsync().ConfigureAwait(false);
                     }
                 }
             }
             catch (Exception ex)
             {
-                _state.Handle(ex);
+                _state.Handle(LiteException.InvalidDatafileState(ex, "DiskWriterQueue failed"));
```
ltetak, line 111: make DiskWriterQueue exceptions "fatal" - these can't be survived and it is best to tear down the entire engine and let the outside observer to restart the engine (possibly with a rebuild)

LiteDB/Engine/Disk/DiskWriterQueue.cs
```diff
@@ -100,14 +102,23 @@ private async Task ExecuteQueue()
...
+            }
+        }
+
+        private void ExhaustQueue()
+        {
+            while (_queue.TryDequeue(out var page))
+            {
+                page.Release();
             }
         }
```
ltetak, line 123: If anything fails here we need to drop this object - this means all pages must be released (including the one that caused the failure - see the finally case below)

Blue101black:
@ltetak were you planning on publishing this PR?

I would love to have a way to hook into the logs from LiteDB to help diagnose issues we are getting in Production.

ltetak:
@Blue101black here it is https://github.com/litedb-org/LiteDB/pull/2576
The rest of this branch makes no sense now. DiskWriterQueue has been completely removed in the master. 👍

PR #2481, opened 2024-05-16
Prevent using disposed snapshots

oleksii-datsiuk:
This fix prevents ENSURE ... Clear() cache error and DB corruption problem, which is described in this issue: https://github.com/mbdavid/LiteDB/issues/2480


JKamsker:
Thanks for the contribution!

PR #2488, opened 2024-06-01
Fix LockRecursionException when doing FindById

### LINKED ISSUE #2471: [BUG] LockRecursionException thrown in very simple use case
**Version**
5.0.19

**Describe the bug**
Here is an MSTest test case that passes, indicating that the Checkpoint() call throws a LockRecursionException with the 'Write lock may not be acquired with read lock held' message that's mentioned in some other bug reports.  This is a simplification of an exception that's being thrown in my application that I don't understand.  Also, I'm a new LiteDB user and could easily have misunderstood something.  Have I forgotten to do something?

This code creates a new database using an expandable MemoryStream, then adds an object to it, receiving the new id.  Then, it looks up the document by this new id receiving that document (a copy).  Finally, the code tries to checkpoint the changes.  At this point, LiteDB throws a LockRecursionException, presumably because it's trying to write changes to the memory stream and a read lock is held by something.  This isn't an iterator that hasn't been allowed to finish, so I'm stumped.

**Code to Reproduce**
        [TestMethod]
        public void TestFragmentDB_FindByIDException()
        {
            var db = new LiteDatabase(new MemoryStream());
            var collection = db.GetCollection<object>("fragtest");

            var fragment = new object { };
            var id = collection.Insert(fragment);

            id.Should().BeGreaterThan(0);

            var frag2 = collection.FindById(id);
            frag2.Should().NotBeNull();

            Action act = () => db.Checkpoint();

            act.Should().Throw<LockRecursionException>();

            db.Dispose();
        }

**Expected behavior**
I expect db.Checkpoint() to not throw an exception.

**Screenshots/Stacktrace**
image
image

**Additional context**
This seems like a very basic use case, so it's most likely that I'm doing something wrong.

If I remove the FindById() call, the checkpoint does not throw the exception.

I'm running this code in the TestExplorer in Visual Studio, which I think is MSTest, but that doesn't seem to be the issue.  If I copy this code to the Program.cs file of my (desktop) application and run it in the main thread before the GUI is created, the code throws the same exception

I believe the test runner configured to run tests sequentially.  Also, I'm running this test alone ... no other tests are running, using LiteDB or otherwise.

JKamsker:
Fixes #2471, https://github.com/mbdavid/LiteDB/issues/2435, https://github.com/mbdavid/LiteDB/issues/2483

Problem was, that the enumerator did not release the readlock when doing FindById. When the Checkpoint then tried to aquire a writelock, the semaphore threw an exception, because there was still a readlock pending.

### DIFF
--- a/LiteDB.Tests/Internals/Extensions_Test.cs
+++ b/LiteDB.Tests/Internals/Extensions_Test.cs
@@ -1,0 +1,44 @@
+﻿using LiteDB.Utils.Extensions;
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using Xunit;
+
+namespace LiteDB.Tests.Internals;
+
+public class Extensions_Test
+{
+    // Asserts that chained IEnumerable<T>.OnDispose(()=> { }) calls the action on dispose, even when chained
+    [Fact]
+    public void EnumerableExtensions_OnDispose()
+    {
+        var disposed = false;
+        var disposed1 = false;
+        var enumerable = new[] { 1, 2, 3 }.OnDispose(() => disposed = true).OnDispose(() => disposed1 = true);
+
+        foreach (var item in enumerable)
+        {
+            // do nothing
+        }
+
+        Assert.True(disposed);
+        Assert.True(disposed1);
+    }
+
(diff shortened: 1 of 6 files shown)


JKamsker:
Ok just found out it was already fixed by https://github.com/mbdavid/LiteDB/pull/2436 but this fix here has a easier git comparison and tests.

PR #2489, opened 2024-06-01
Add support for on-the-fly document upgrades

JKamsker:
This change would allow me to improve my library (https://github.com/JKamsker/LiteDb.Migration) significantly. The documents would then only be upgrades on load and not all at once.

### DIFF
--- a/LiteDB/Engine/EngineSettings.cs
+++ b/LiteDB/Engine/EngineSettings.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
+
 using static LiteDB.Constants;
 
 namespace LiteDB.Engine
@@ -65,6 +66,11 @@ namespace LiteDB.Engine
         /// If detect it's a older version (v4) do upgrade in datafile to new v5. A backup file will be keeped in same directory
         /// </summary>
         public bool Upgrade { get; set; } = false;
+
+        /// <summary>
+        /// Is used to transform a <see cref="BsonValue"/> from the database on read. This can be used to upgrade data from older versions.
+        /// </summary>
+        public Func<string, BsonValue, BsonValue> ReadTransform { get; set; }
 
         /// <summary>
         /// Create new IStreamFactory for datafile
--- a/LiteDB/Engine/EngineState.cs
+++ b/LiteDB/Engine/EngineState.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -8,7 +8,6 @@ using System.Text;
 using System.Threading.Tasks;
 
 using static LiteDB.Constants;
-
 
 namespace LiteDB.Engine
 {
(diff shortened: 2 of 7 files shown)


LiteDB/Client/Database/LiteDatabase.cs
```diff
@@ -32,19 +32,19 @@ public partial class LiteDatabase : ILiteDatabase
         /// <summary>
         /// Starts LiteDB database using a connection string for file system database
         /// </summary>
-        public LiteDatabase(string connectionString, BsonMapper mapper = null)
-            : this(new ConnectionString(connectionString), mapper)
+        public LiteDatabase(string connectionString, BsonMapper mapper = null, Action<EngineSettings> engineSettingsAction = null)
```
pictos, line 35: this is a breaking change, so in order to release it in the SR I suggest to add an overload to this ctor accepting this new Action

LiteDB/Engine/EngineSettings.cs
```diff
@@ -66,6 +66,11 @@ public class EngineSettings
         /// </summary>
         public bool Upgrade { get; set; } = false;
 
+        /// <summary>
+        /// Is used to transform a bsonvalue from the database on read. This can be used to upgrade data from older versions.
```
pictos, line 70:
```suggestion
         /// Is used to transform a <see cref="BsonValue"/> from the database on read. This can be used to upgrade data from older versions.
```

LiteDB/Engine/EngineState.cs
```diff
@@ -51,5 +51,12 @@ public bool Handle(Exception ex)
 
             return true;
         }
+
+        public BsonValue ReadTransform(string collection, BsonValue value)
+        {
+            if (_settings?.ReadTransform == null) return value;
```
pictos, line 57: Using pattern-matching to check nullability is safer and better.

```suggestion
            if (_settings?.ReadTransform is null) return value;
```

pictos (commented):
Looking into the issue for more context, I think is better to @mbdavid to take a look on this and see if he wants to perform this change, or if using the `protected` methods are enough

JKamsker:
@pictos Thanks for the review, how did that slip through, that its a breaking change 😅 
Reverted the breaking changes and updated your requested changes.

So in my latest commit i added a deserialization hook.

Upsides:
- It is easier to add your own hook into that, because you only need a reference to the mapper
- Should be easier on the cpu because it only transforms those documents, which are actually needed.

Downsides:
- Queries may not run as expected, because there could be a transform between the query engine and the actual transform. But if that is a concern, the user should just use the ReadTransform in the engine settings.

LiteDB/LiteDB.csproj
```diff
@@ -28,6 +28,7 @@
     <SignAssembly Condition="'$(OS)'=='Windows_NT'">true</SignAssembly>
     <AssemblyOriginatorKeyFile Condition="'$(Configuration)' == 'Release'">LiteDB.snk</AssemblyOriginatorKeyFile>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <LangVersion>9.0</LangVersion>
```
pictos, line 31:
```suggestion
	<LangVersion>latest</LangVersion>
```

pictos (commented):
I'm away from my pc, so I can't test it right now

JKamsker:
Some tests failed before which are succeding (not the tests i touched)...sooo...

@pictos Can we merge? 😅

mbdavid (approved):

PR #2497, opened 2024-06-06
Add failing tests

JKamsker:
That way people could submit prs only containing tests that should work but currently don't

### DIFF
--- a/LiteDB.Tests/Failing/Issue2475_tests.cs
+++ b/LiteDB.Tests/Failing/Issue2475_tests.cs
@@ -1,0 +1,41 @@
+﻿using LiteDB.Tests.Utils;
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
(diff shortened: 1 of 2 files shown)


pictos:
But why? I would say that's better to contributors implementing the test with the actual fix... I believe that kind of test will pollute the test suite

JKamsker:
This would be a benchmark to measure the (hopefully increasing) stability of the library

pictos:
> This would be a benchmark to measure the (hopefully increasing) stability of the library.

can you expand more on this?

> That way we know what the user would expect and we or someone else can fix it if they got the resources for it.

on GitHub we already have `issues` to mark what needs to be worked, and the community interaction on those issues will point us to the highest priority ones. So how having a bunch of failing tests would be a better measurement system than what GitHub already provide to us?

pictos:
can we close this one?

PR #2498, opened 2024-06-06
Bugs/fix 2487 allow contains whitespaces

### LINKED ISSUE #2487: Incorrect argument validation exception
Query.Contains checks the String.IsNullOrWhiteSpace instead of String.IsNullOrEmpty

https://github.com/mbdavid/LiteDB/blob/5a4696606b4fc58554f01d725f621bd3e553fd18/LiteDB/Client/Structures/Query.cs#L126

1. It causes incorrect ArgumentNullExcpeption for a spaces sequence
2. There is no reason to restrict the search by spaces, because the same LIKE query works correct

JKamsker:
Fixes #2487 and contains a minor perf improvement

### DIFF
--- a/LiteDB.Tests/Issues/Issue2487_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2487_Tests.cs
@@ -1,5 +1,7 @@
 ﻿using FluentAssertions;
 
+using System.Diagnostics;
+
 using Xunit;
 
 namespace LiteDB.Tests.Issues;
@@ -27,6 +29,8 @@ public class Issue2487_tests
         collection.Insert(new DataClass { Foo = "bar", Bar = "abc" });
         collection.Insert(new DataClass { Foo = " ", Bar = "def" });
         collection.Insert(new DataClass { Foo = "fo bar", Bar = "def" });
+        collection.Insert(new DataClass { Foo = "", Bar = "def" });
+        collection.Insert(new DataClass { Foo = null, Bar = "def" });
 
         var containsAction = () => collection.FindOne(x => x.Foo.Contains(" "));
         containsAction.Should().NotThrow();
--- a/LiteDB.Tests/Issues/Issue2487_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2487_Tests.cs
@@ -1,0 +1,41 @@
+﻿using FluentAssertions;
+
+using Xunit;
+
+namespace LiteDB.Tests.Issues;
+
+public class Issue2487_tests
+{
+    private class DataClass
+    {
+        [BsonId]
+        public int Id { get; set; }
+
+        public string Foo { get; set; }
+
+        public string Bar { get; set; }
+    }
+
+    [Fact]
+    public void Test_Contains_EmptyStrings()
+    {
+        using var engine = new ConnectionString(":memory:").CreateEngine();
+
+        using var db = new LiteDatabase(engine);
+        var collection = db.GetCollection<DataClass>("data");
+
+        collection.Insert(new DataClass { Foo = "bar", Bar = "abc" });
+        collection.Insert(new DataClass { Foo = " ", Bar = "def" });
+        collection.Insert(new DataClass { Foo = "fo bar", Bar = "def" });
+
+        var containsAction = () => collection.FindOne(x => x.Foo.Contains(" "));
+        containsAction.Should().NotThrow();
+
+        var def = containsAction();
+        def.Should().NotBeNull();
+        def.Bar.Should().Be("def");
+
+        var shouldExecute = () => engine.Query("data", Query.All(Query.Contains("Foo", " ")));
+        shouldExecute.Should().NotThrow();
+    }
+}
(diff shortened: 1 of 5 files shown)


LiteDB/Utils/Extensions/StringExtensions.cs
```diff
@@ -1,16 +1,25 @@
...
 
 namespace LiteDB
 {
     internal static class StringExtensions
     {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsNullOrWhiteSpace(this string str)
         {
-            return str == null || str.Trim().Length == 0;
+            return string.IsNullOrWhiteSpace(str);
```
pictos, line 16: Instead of calling the existing method, isn't better to replace all calls by this?

JKamsker: The clr should inline it without aggressiveinlining, but i added it nontheless. https://web.archive.org/web/20120210222409/http://blogs.microsoft.co.il/blogs/sasha/archive/2012/01/20/aggressive-inlining-in-the-clr-4-5-jit.aspx

JKamsker: grafik
grafik

Here we see the content of IsNullOrWhiteSpace exists 2x: One in IsNullOrWhiteSpace and one in main
grafik
grafik

pictos: I'm aware about the AgressiveInling, but isn't guarantee that the compiler will inline it in all scenarios.
My question is more why keeping a method that will just forward to some method that exists on BCL, I'm fine to adding those but will be hard to point all devs to this extension methods

LiteDB.Tests/Issues/Issue2487_Tests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    [Fact]
+    public void Test_Contains_EmptyStrings()
+    {
+        using var engine = new ConnectionString(":memory:").CreateEngine();
+
+        using var db = new LiteDatabase(engine);
+        var collection = db.GetCollection<DataClass>("data");
+
+        collection.Insert(new DataClass { Foo = "bar", Bar = "abc" });
```
pictos, line 27: can you add a case where the value is and `string.Empty` as well? Then I think it should be good to be merged

LiteDB/Utils/Extensions/StringExtensions.cs
JKamsker: I understand your point about the redundancy of having an extension method that essentially forwards to a BCL method. 
However, I believe there are a couple of reasons to keep our custom extension method:

- The extension method provides a level of convenience and readability for our codebase. It's a familiar and established pattern in the repository and helps maintain consistency.

- Changing all 50 or so calls to the method would introduce unnecessary churn and pollute the git history. It's better to keep the existing method and avoid making widespread changes for minimal benefit. I, atleast do not want to pollute this PR with that.

Additionally, I find the "".IsNullOrWhiteSpace() extension method more readable and it follows OOP principles, even if it's not strictly OOP. It makes the code more intuitive and aligns with the way we think about object-oriented design.

Anyway, i have no hard feelings if you want to change it, i would suggest we do it in a separate PR. I, for my part prefer a clean PR 😅

LiteDB.Tests/Issues/Issue2487_Tests.cs
JKamsker: You mean that one DataClass.Foo is string.Empty?

pictos: Yes, having values that are, null; empty

LiteDB/Utils/Extensions/StringExtensions.cs
pictos: As I said I'm fine to keeping this extension method for now. Even not seeing too much value on it.

> even if it's not strictly OOP

me neither (:

pictos:
@JKamsker we do have a failing build, can you verify the reason?

JKamsker:
> @JKamsker we do have a failing build, can you verify the reason?

Debug builds just fine, release doesnt.
Because: https://github.com/mbdavid/LiteDB/blob/610e530bb2d70be9718777dff0ebaba66a3d9cdd/LiteDB/Utils/Constants.cs#L7-L10
And ``CreateEngine`` in https://github.com/mbdavid/LiteDB/blob/fad6af353d86514faa90e46e3db1efb73369331e/LiteDB.Tests/Issues/Issue2487_Tests.cs#L22

Which is internal
https://github.com/mbdavid/LiteDB/blob/610e530bb2d70be9718777dff0ebaba66a3d9cdd/LiteDB/Client/Structures/ConnectionString.cs#L110

@pictos  Do you think its safe to remove the ``#if DEBUG`` in ``Constants`` ? Otherwise we cant test release builds. I Have no idea whats the benefit of ``InternalsVisibleTo`` when you cant use it half of the time.

LiteDB/Utils/Extensions/StringExtensions.cs
Dainius14: IMO better replace all with the standard, because otherwise there will be both `string.IsNullOrWhiteSpace(str)` and `str.IsNullOrWhiteSpace()` calls in the codebase, which would be just a small inconsistency for new contributors, without a clear indication which one is preferred

mbdavid (approved):

PR #2499, opened 2024-06-07
Avoid using AesStream class when opening an older version of LiteDB d…

jdtkw:
…atabase

### DIFF
--- a/LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
+++ b/LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
@@ -17,13 +17,15 @@ namespace LiteDB.Engine
         private readonly string _password;
         private readonly bool _readonly;
         private readonly bool _hidden;
+		private readonly bool _useAesStream;
 
-        public FileStreamFactory(string filename, string password, bool readOnly, bool hidden)
+		public FileStreamFactory(string filename, string password, bool readOnly, bool hidden, bool useAesStream = true)
         {
             _filename = filename;
             _password = password;
             _readonly = readOnly;
             _hidden = hidden;
+            _useAesStream = useAesStream;
         }
 
         /// <summary>
@@ -57,7 +59,7 @@ namespace LiteDB.Engine
                 File.SetAttributes(_filename, FileAttributes.Hidden);
             }
 
-            return _password == null ? (Stream)stream : new AesStream(_password, stream);
+            return _password == null || !_useAesStream ? (Stream)stream : new AesStream(_password, stream);
         }
 
         /// <summary>
--- a/LiteDB/Engine/Services/RebuildService.cs
+++ b/LiteDB/Engine/Services/RebuildService.cs
@@ -22,14 +22,19 @@ namespace LiteDB.Engine
         {
             _settings = settings;
 
-            // open, read first 16kb, and close data file
-            var buffer = this.ReadFirstBytes();
+			// test for prior version
+			var bufferV7 = this.ReadFirstBytes(false);
+			if (FileReaderV7.IsVersion(bufferV7))
+            {
+                _fileVersion = 7;
+                return;
+            }
 
-            // test for valid reader to use
-            _fileVersion = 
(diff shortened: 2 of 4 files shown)


JKamsker:
Can you add tests that show that the upgrade works with your changes?

LiteDB/Engine/Services/RebuildService.cs
```diff
@@ -22,14 +22,19 @@ public RebuildService(EngineSettings settings)
         {
             _settings = settings;
 
-            // open, read first 16kb, and close data file
-            var buffer = this.ReadFirstBytes();
+			// test for prior version
```
pictos, line 25: please change the indentation to be space instead of tabs

LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
```diff
@@ -17,13 +17,15 @@ internal class FileStreamFactory : IStreamFactory
         private readonly string _password;
         private readonly bool _readonly;
         private readonly bool _hidden;
+		private readonly bool _useAesStream;
 
-        public FileStreamFactory(string filename, string password, bool readOnly, bool hidden)
+		public FileStreamFactory(string filename, string password, bool readOnly, bool hidden, bool useAesStream = true)
```
pictos, line 22: This is a breaking change, instead of adding one more parameter on the ctor, please create a new one with this parameter, and use `this` if possible to forward to the old one

LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
```diff
@@ -17,13 +17,15 @@ internal class FileStreamFactory : IStreamFactory
         private readonly string _password;
         private readonly bool _readonly;
         private readonly bool _hidden;
+		private readonly bool _useAesStream;
```
pictos, line 20: please change the indentation to be space instead of tabs

pictos (commented):
Thanks for your contribution, I've a minor comment to avoid this be a breaking change and with that we should be able to release it as a SR

pictos (approved):
Thanks for your contribution, I've a minor comment to avoid this be a breaking change and with that we should be able to release it as a SR

pictos (changes requested):
Wrong button

LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
JKamsker: The class is internal. No breaking.

JKamsker:
The code feels a bit odd to me but it works. I could add my tests if you allow additions by maintainers.

LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
pictos: Ops, my bad

JKamsker (approved):
lgtm

LiteDB/Engine/Disk/StreamFactory/FileStreamFactory.cs
jdtkw: Done.

jdtkw:
> The code feels a bit odd to me but it works. I could add my tests if you allow additions by maintainers.



> The code feels a bit odd to me but it works. I could add my tests if you allow additions by maintainers.

Yeah, I went for a fix for the problem that would limit any form of adverse interaction with other code / use cases. Not 100% happy with it myself, but feel it's not going to affect normal operation once it has upgraded to to the latest format.

LiteDB/Engine/Services/RebuildService.cs
jdtkw: Done.

jdtkw:
> Can you add tests that show that the upgrade works with your changes?

Thanks for doing this.

pictos (approved):

pictos:
@jdtkw we will see other PRs to merge and then schedule a SR version, that should not take so long.

Thanks for your contribution

PR #2508, opened 2024-06-12
[♻️Housekeeping] Remove BufferPool class and use .net implementation

pictos:
There're a lot of places where Litedb news up an array just to discard it later, this is bad for allocations and perf. So:

- I added the the ArrayPool provided by the .NET team
- Removed old Pool implementations and replace then by .NET's one
- Change, where is possible, in the code places that news up an array to discard then later to use the ArrayPool


LiteDB/Engine/Disk/Serializer/BufferReader.cs
```diff
@@ -161,13 +163,13 @@ public string ReadString(int count)
             else
             {
                 // rent a buffer to be re-usable
-                var buffer = BufferPool.Rent(count);
```
JKamsker, line 164: Does using a [BufferOwner](https://gist.github.com/JKamsker/27f65465d401b03e68f3822406227285) make sense here?

JKamsker (commented):
Generally, i would love to see this step to finally happen.

- Why do we not centralize the buffer pool declaration or just use ArrayPool<byte> directly
- Is the additional dependency also required in newer versions? Will it be installed in net8? The ArrayPool in net8 is in ``System.Runtime``

LiteDB/Engine/Disk/Serializer/BufferReader.cs
```diff
@@ -251,13 +253,13 @@ private T ReadNumber<T>(Func<byte[], int, T> convert, int size)
             }
             else
             {
-                var buffer = BufferPool.Rent(size);
+                var buffer = bufferPool.Rent(size);
```
JKamsker, line 256: **Warning!** We need to to zero the loaned buffer! Otherwise the convert will return unexpected results! Rel: [ArrayPool<T>.Rent(Int32) Method](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.rent?view=net-8.0&devlangs=csharp&f1url=%3FappId%3DDev17IDEF1%26l%3DEN-US%26k%3Dk(System.Buffers.ArrayPool%25601.Rent)%3Bk(DevLang-csharp)%26rd%3Dtrue#remarks)

LiteDB/Engine/Disk/Serializer/BufferReader.cs
```diff
@@ -328,13 +330,13 @@ public ObjectId ReadObjectId()
             }
             else
             {
-                var buffer = BufferPool.Rent(12);
+                var buffer = bufferPool.Rent(12);
```
JKamsker, line 333: **Warning!** We need to to zero the loaned buffer! Otherwise the ObjectId could return unexpected results! Rel: [ArrayPool<T>.Rent(Int32) Method](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1.rent?view=net-8.0&devlangs=csharp&f1url=%3FappId%3DDev17IDEF1%26l%3DEN-US%26k%3Dk(System.Buffers.ArrayPool%25601.Rent)%3Bk(DevLang-csharp)%26rd%3Dtrue#remarks)

LiteDB/Engine/Disk/Streams/AesStream.cs
```diff
@@ -52,6 +55,9 @@ public AesStream(string password, Stream stream)
             // start stream from zero position
             _stream.Position = 0;
 
+            var checkBuffer = bufferPool.Rent(32);
```
JKamsker, line 58: ``checkBuffer.Length`` will be used. Since the required 32 bytes is a minimum, the byte array returned could be larger.
Please adjust the code that uses the checkbuffer.

LiteDB/Engine/Engine/Upgrade.cs
```diff
@@ -19,20 +23,21 @@ private void TryUpgrade()
...
                 _settings.Filename,
                 FileMode.Open,
                 FileAccess.Read,
                 FileShare.Read, 1024))
             {
-                var buffer = new byte[1024];
+
 
                 stream.Position = 0;
                 stream.Read(buffer, 0, buffer.Length);
```
JKamsker, line 36: the length is not 1024 anymore! It could be 10 mb depending on what the pool has left.

JKamsker (changes requested):

LiteDB/Engine/Disk/Serializer/BufferReader.cs
pictos: no, because it will be boxed and will allocate, and here I'm trying to avoid allocations. When we bump the TFM we may use `ref struct` and better APIs

LiteDB/Engine/Disk/Streams/AesStream.cs
pictos: which test should fail if the array's length isn't 32bytes?

mbdavid (approved):
well done, thanks

Issue #2505, opened 2024-06-10
Make all internal Task SyncContext agnostic


pictos:
Raised on #2447, the long running task is sensitive to `SyncronizationContext`, that's devs can force that code execute inside the `SyncronizationContext`, and probably isn't what we want.

So in order to fix that we need to make sure our tasks will run under the `TaskScheduler.Default`.

The linked PR is one example, but we should investigate on our code base more places where we may have this issue.

alexbereznikov:
@pictos Probably this issue should be closed due to #2515

PR #2511, opened 2024-06-13
Make all internal Task SyncContext agnostic

alexbereznikov:
This PR addresses #2505

In fact, the only place in LiteDB where we're firing up a new task is in DiskWriterQueue.

Also, changed write/flush methods to async ones.

### DIFF
--- a/LiteDB/Engine/Disk/DiskWriterQueue.cs
+++ b/LiteDB/Engine/Disk/DiskWriterQueue.cs
@@ -56,10 +56,7 @@ namespace LiteDB.Engine
 
             lock (_queueSync)
             {
-                if (_task == null)
-                {
-                    _task = Task.Factory.StartNew(ExecuteQueue, TaskCreationOptions.LongRunning);
-                }
+                _task ??= Task.Run(ExecuteQueue);
             }
         }
 
@@ -76,7 +73,7 @@ namespace LiteDB.Engine
         /// <summary>
         /// Execute all items in queue sync
         /// </summary>
-        private void ExecuteQueue()
+        private async Task ExecuteQueue()
         {
             try
             {
@@ -84,7 +81,7 @@ namespace LiteDB.Engine
                 {
                     if (_queue.TryDequeue(out var page))
                     {
-                        WritePageToStream(page);
+                        await WritePageToStream(page).ConfigureAwait(false);
                     }
                     else
                     {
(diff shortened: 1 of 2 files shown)


LiteDB/Engine/Disk/DiskWriterQueue.cs
```diff
@@ -76,15 +73,15 @@ public void Wait()
...
+        private async Task ExecuteQueue()
         {
             try
             {
                 while (true)
                 {
                     if (_queue.TryDequeue(out var page))
                     {
-                        WritePageToStream(page);
+                        await WritePageToStream(page).ConfigureAwait(false);
```
pictos, line 84: Please, remove this and other `ConfigureAwait`s, they aren't doing not here, just polluting the code

LiteDB/Utils/Extensions/StreamExtensions.cs
```diff
@@ -20,5 +20,19 @@ public static void FlushToDisk(this Stream stream)
                 stream.Flush();
             }
         }
+
+        public static async Task FlushToDiskAsync(this Stream stream)
```
pictos, line 24: why you created this extension method? Is doing nothing useful...

pictos (changes requested):

pictos (changes requested):
Thanks for the PR. I've some comments around the code

LiteDB/Engine/Disk/DiskWriterQueue.cs
alexbereznikov: Removed

LiteDB/Utils/Extensions/StreamExtensions.cs
alexbereznikov: Removed, thanks

LiteDB/Engine/Disk/DiskWriterQueue.cs
```diff
@@ -56,10 +56,7 @@ public void EnqueuePage(PageBuffer page)
 
             lock (_queueSync)
             {
-                if (_task == null)
-                {
-                    _task = Task.Factory.StartNew(ExecuteQueue, TaskCreationOptions.LongRunning);
-                }
+                _task ??= Task.Run(ExecuteQueue);
```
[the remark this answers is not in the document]
alexbereznikov: Please clarify what do you mean :) 

Without tests, there are only 2 possible instantiations of `DiskService` - both are inside `LiteEngine.Open` method. One of them gets executed only during auto rebuild, and the previous instance is disposed before that.

pictos:
Closing this on in favor of #2515. Thanks for this PR

PR #2515, opened 2024-06-26
Remove DiskWriterQueue to avoid adicional thread to write on disk

mbdavid:


pictos (approved):

LiteDB/Engine/Disk/DiskService.cs
```diff
@@ -171,40 +162,55 @@ public PageBuffer NewPage()
...
         /// </summary>
-        public int WriteAsync(IEnumerable<PageBuffer> pages)
+        public int WriteLogDisk(IEnumerable<PageBuffer> pages)
         {
             var count = 0;
+            var stream = _writer.Value;
 
-            foreach (var page in pages)
+            // do a global write lock - only 1 thread can write on disk at time
+            lock(stream)
```
alexbereznikov, line 173: A minor thing - should be ok here, but usually it is better to have a separate object for locking

alexbereznikov (approved):

Issue #2536, opened 2024-08-30
[BUG] Race condition in BsonMapper


einarmo:
**Version**
LiteDB 5.0.21
.NET 8.0.401
Linux (Fedora 38)

**Describe the bug**
Specifically, what I see is an occasional stacktrace:

```
System.NotSupportedException : Member Id not found on BsonMapper for type MyNamespace.MyType
 Stack Trace:
at LiteDB.LinqExpressionVisitor.ResolveMember(MemberInfo member)
at LiteDB.LinqExpressionVisitor.VisitMember(MemberExpression node)
at LiteDB.LinqExpressionVisitor.ResolvePattern(String pattern, Expression obj, IList`1 args)
at LiteDB.LinqExpressionVisitor.VisitMethodCall(MethodCallExpression node)
at System.Linq.Expressions.ExpressionVisitor.VisitLambda[T](Expression`1 node)
at LiteDB.LinqExpressionVisitor.VisitLambda[T](Expression`1 node)
at LiteDB.LinqExpressionVisitor.Resolve(Boolean predicate)
at LiteDB.BsonMapper.GetExpression[T,K](Expression`1 predicate)
at LiteDB.LiteCollection`1.DeleteMany(Expression`1 predicate)
```

when running an expression like

```c#
collection.DeleteMany(item => idsToDelete.Contains(item.Id))
```

**Code to Reproduce**
I cannot reproduce this consistently, it only happens once in a while, likely while I am running multiple concurrent transactions not against the same collection, but using the same types.

I believe this is a race condition caused by #2493, here's the possible race I see that this PR introduced:

 - Thread A enters `GetEntityMapper` [here](https://github.com/mbdavid/LiteDB/blame/391cc9318c5be6e56cb71b3ce14b1ed9cb324763/LiteDB/Client/Mapper/BsonMapper.cs#L238), and does not find the mapper.
 - Thread A calls `BuildAddEntityMapper`, and goes past [this line](https://github.com/mbdavid/LiteDB/blame/391cc9318c5be6e56cb71b3ce14b1ed9cb324763/LiteDB/Client/Mapper/BsonMapper.cs#L257), adding an entity mapper to the _entities dictionary.
 - Thread A is preempted by thread B, which enters `GetEntityMapper`.
 - Thread B _finds_ a mapper and returns, but this mapper won't work, as it has not yet been populated with fields, resulting in the stacktrace above.

[...]

PR #2537, opened 2024-09-03
Fix race condition in #2536

einarmo:
This fixes a race condition in BsonMapper (#2536), caused by a fix to a different issue in #2493.

It seems like the current approach of checking the dictionary twice is deliberate, likely for performance reasons. That said, I don't believe reading from a dictionary that may be in the process of being updated is actually safe to begin with.

If this is unacceptable perf-wise I believe the locking must be swapped with a proper read-write lock.


einarmo:
Closing in favor of #2538

PR #2538, opened 2024-09-06
#2536 Fix race condition in bsonmapper

### LINKED ISSUE #2536: [BUG] Race condition in BsonMapper

JKamsker:
Fixes https://github.com/mbdavid/LiteDB/issues/2536

### DIFF
--- a/LiteDB/Client/Mapper/BsonMapper.cs
+++ b/LiteDB/Client/Mapper/BsonMapper.cs
@@ -30,7 +30,12 @@ namespace LiteDB
         /// Mapping cache between Class/BsonDocument
         /// </summary>
         private readonly Dictionary<Type, EntityMapper> _entities = new Dictionary<Type, EntityMapper>();
-
+        
+        /// <summary>
+        /// Unfinished mapping cache between Class/BsonDocument
+        /// </summary>
+        private readonly Dictionary<Type, EntityMapper> _buildEntities = new Dictionary<Type, EntityMapper>();
+        
         /// <summary>
         /// Map serializer/deserialize for custom types
         /// </summary>
@@ -227,6 +232,8 @@ namespace LiteDB
         #endregion
 
         #region GetEntityMapper
+        
+        
 
         /// <summary>
         /// Get property mapper between typed .NET class and BsonDocument - Cache results
@@ -235,13 +242,30 @@ namespace LiteDB
         {
             //TODO: needs check if Type if BsonDocument? Returns empty EntityMapper?
 
-            if (!_entities.TryGetValue(type, out EntityMapper mapper))
+            if (_entities.TryGetValue(type, out EntityMapper mapper))
             {
-                lock (_entities)
+                return mapper;
+            }
+
+            lock (_entities)
+            {
+                if (_entities.TryGetValue(type, out mapper))
                 {
-                    if (!_entities.TryGetValue(type, out mapper))
-                        return this.BuildAddEntityMapper(type);
+                    return mapper;
                 }
+
+                if(_buildEntities.TryGetValue(type, out EntityMapper buildMapper))
+                {
+                    return buildMapper;
+                }
+                        
+                var newMapper = new EntityMapper(type);
+                _buildEntities[type] = newMapper;
+                this.BuildEntityMapper(newMapper);
+                _entities[type] = newMapper;
+                        
+                _buildEntities.Remove(type);
+                return newMapper;
             }
 
             return mapper;
@@ -251,17 +275,17 @@ namespace LiteDB
         /// Use this method to override how your class can be, by default, mapped from entity to Bson document.
         /// Returns an EntityMapper from each requested Type
         /// </summary>
-        protected virtual EntityMapper BuildAddEntityMapper(Type type)
+        protected void BuildEntityMapper(EntityMapper mapper)
         {
-            var mapper = new EntityMapper(type);
-            _entities[type] = mapper;//direct add into entities, to solove the DBRef [ GetEntityMapper > BuildAddEntityMapper > RegisterDbRef > RegisterDbRefItem > GetEntityMapper ] Loop call recursion,we stoped at here and GetEntityMapper's _entities.TryGetValue
+            // var mapper = new EntityMapper(type);
+            // _entities[type] = mapper;//direct add into entities, to solove the DBRef [ GetEntityMapper > BuildAddEntityMapper > RegisterDbRef > RegisterDbRefItem > GetEntityMapper ] Loop call recursion,we stoped at here and GetEntityMapper's _entities.TryGetValue
 
             var idAttr = typeof(BsonIdAttribute);
             var ignoreAttr = typeof(BsonIgnoreAttribute);
             var fieldAttr = typeof(BsonFieldAttribute);
             var dbrefAttr = typeof(BsonRefAttribute);
 
-            var members = this.GetTypeMembers(type);
+            var members = this.GetTypeMembers(mapper.ForType);
             var id = this.GetIdMember(members);
 
             foreach (var memberInfo in members)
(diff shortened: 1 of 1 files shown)


LiteDB/Client/Mapper/BsonMapper.cs
```diff
@@ -235,13 +242,30 @@ internal EntityMapper GetEntityMapper(Type type)
...
             {
-                lock (_entities)
+                if (_entities.TryGetValue(type, out mapper))
                 {
-                    if (!_entities.TryGetValue(type, out mapper))
-                        return this.BuildAddEntityMapper(type);
+                    return mapper;
                 }
+
+                if(_buildEntities.TryGetValue(type, out EntityMapper buildMapper))
```
einarmo, line 257: I think this case is impossible, and shouldn't be necessary to check.

einarmo (commented):
I think this makes sense, though the whole thing is IMO a little unpleasant.

Probably better than my solution in terms of performance.

One thing I'm still unsure of is whether it is actually safe to concurrently insert into a dictionary while reading from it on another thread, I think there is a chance it fails, if the underlying storage needs to grow, but I'm not familiar enough with the details of Dictionary to say for sure.

SkitX1337 (approved):

LiteDB/Client/Mapper/BsonMapper.cs
```diff
@@ -30,7 +30,12 @@ public partial class BsonMapper
         /// Mapping cache between Class/BsonDocument
         /// </summary>
         private readonly Dictionary<Type, EntityMapper> _entities = new Dictionary<Type, EntityMapper>();
-
+        
+        /// <summary>
+        /// Unfinished mapping cache between Class/BsonDocument
+        /// </summary>
+        private readonly Dictionary<Type, EntityMapper> _buildEntities = new Dictionary<Type, EntityMapper>();
```
pictos, line 37: how about using ConcurrentDictionary?

JKamsker: We don't need concurrent access to this because it is only ever used in building the type mapping, which is always happening in the same thread.

LiteDB/Client/Mapper/BsonMapper.cs
```diff
@@ -235,13 +242,30 @@ internal EntityMapper GetEntityMapper(Type type)
         {
             //TODO: needs check if Type if BsonDocument? Returns empty EntityMapper?
 
-            if (!_entities.TryGetValue(type, out EntityMapper mapper))
+            if (_entities.TryGetValue(type, out EntityMapper mapper))
+            {
+                return mapper;
+            }
+
+            lock (_entities)
```
pictos, line 250: You mentioned that is not needed to use `ConcurrentDictionary`, but we need a lock here in order to access and update its values (for both), are you sure isn't better to use that collection and avoid the `lock` here?

JKamsker: In the best case (which happens 99% of the time), we don`t lock. Only initially, when first creating the mapping, we need the lock. 

In addidtion to that, this behaviour is the same as the original code before the [bugfix that caused this bug](https://github.com/mbdavid/LiteDB/pull/2493)

pictos (approved):

LiteDB/Client/Mapper/BsonMapper.cs
```diff
@@ -30,7 +30,12 @@ public partial class BsonMapper
         /// Mapping cache between Class/BsonDocument
         /// </summary>
         private readonly Dictionary<Type, EntityMapper> _entities = new Dictionary<Type, EntityMapper>();
```
vonzshik, line 32: I'm not sure it's a good idea to allow concurrent access to `Dictionary`, given that [it's not thread-safe](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-8.0#thread-safety) and there is possible concurrent read and write happening.

JKamsker: https://github.com/mbdavid/LiteDB/pull/2538/files/4bdb4b700d0ed60e72656c84dc674a2c6004682e#r1778334038

JKamsker: No. The original issue was actually due to recursive access. That's a difference.

vonzshik: I'm talking about https://github.com/mbdavid/LiteDB/issues/2536. It's possible to have multiple threads doing concurrent read (which is fine) but at the same time one thread will be writing to that dictionary, and this situation isn't thread safe.

JKamsker: Any access to _buildEntities is guarded by a lock

https://github.com/JKamsker/LiteDB/blob/4bdb4b700d0ed60e72656c84dc674a2c6004682e/LiteDB/Client/Mapper/BsonMapper.cs#L250-L269

vonzshik: I'm not talking about `_buildEntities` but `_entities`

JKamsker: The double-checked locking has always been there and i cant find any issues concerning this regard. https://github.com/mbdavid/LiteDB/pull/2493/files#diff-db4637fbacd6bcf21b8bc9df379f084f98e4cac70c3d9858bc8df4d28bdbe7efL234

vonzshik: Here you go #2563, repro which throws nre

PR #2551, opened 2024-10-04
update net8 target to net8.0

WeihanLi:
image

https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#what-about-net-10

`net8` is not a standard target framework name which should not be used, update to the standard `net8.0`


dgerding:
I was about to make the same PR. I rebuilt solution in .net 8 and all tests pass. Don't really follow the ".net 10" reference, but would like to work against later LTR for .net core. Theoretically faster, too? Thanks!

Joy-less (approved):
Looks good to me. All tests passed.

PR #2552, opened 2024-10-04
Fix #2534

### LINKED ISSUE #2534: [BUG] iterating over an  IEnumerable<LiteFileInfo<TFileId>> errors out on the second element if the DB is opened in shared mode
For a test, I changed the connection mode for LiteDb from exclusive to shared. And then, operations on files started erroring out.

the code I'm using looks like this:

```
var fs = db.GetStorage<string>("audioFiles");
var files = fs.Find(u => u.Id.StartsWith(path));
foreach (var file in files)
{ // do something with file
}
```

this works fine if the connection mode is direct. If you then change the connection mode to shared, It fails when there's multiple files (at least two) - the first iteration works, on the second iteration gets you a `LiteException `with message 

> This engine instance already disposed

If you get a List instead of an Enumerable (`var files = fs.Find(u => u.Id.StartsWith(path)).ToList();`) it works even in shared mode.

I would expect the iteration to work in shared mode, too.

LiteDB version is 5.0.21 on .NET 8.0.7 on W11 Pro 2023H2

Joy-less:
This fixes issue #2534 where updating a collection during an iteration fails in shared mode.

Previously, functions like `Collection.Update()` would always open and close the database, causing a dispose error when recursion was used. Now, it only closes the database if it was responsible for opening it.

I'm not sure that this is the best way to solve the problem. It might be a good idea to open a transaction whenever iterating a collection instead. I would have expected this issue to be more impactful.

### DIFF
--- a/LiteDB.Tests/Engine/Recursion_Tests.cs
+++ b/LiteDB.Tests/Engine/Recursion_Tests.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using Xunit;
+
+namespace LiteDB.Tests.Engine;
+
+public class Recursion_Tests {
+    [Fact]
+    public void UpdateInFindAll()
+    {
+        Test(collection =>
+        {
+            foreach (BsonDocument document in collection.FindAll())
+            {
+                collection.Update(document);
+            }
+        });
+    }
+    [Fact]
+    public void InsertDeleteInFindAll()
+    {
+        Test(collection =>
+        {
+            foreach (BsonDocument document in collection.FindAll())
+            {
+                BsonValue id = collection.Insert(new BsonDocument());
+                collection.Delete(id);
+            }
+        });
+    }
+    [Fact]
+    public void QueryInFindAll()
+    {
+        Test(collection =>
+        {
+            foreach (BsonDocument document in collection.FindAll())
+            {
+                collection.Query().Count();
+            }
+        });
+    }
+
+    private void Test(Action<ILiteCollection<BsonDocument>> action)
+    {
+        using LiteDatabase database = new(new ConnectionString()
+        {
+            Filename = "Demo.db",
+            Connection = ConnectionType.Shared,
+        });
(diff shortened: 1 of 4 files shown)


pictos:
Hey @Joy-less thanks for your fix, do you mind to add a recursion unit test as well? Since you described it on the PR description I think would be great to validate that this changes fix it and avoid regressions in future

Joy-less:
@pictos I added the recursion tests. I also updated the test packages which were way out-of-date.

LiteDB.Tests/Engine/Recursion_Tests.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using Xunit;
+
+namespace LiteDB.Tests.Engine;
+
+public class Recursion_Tests {
+    [Fact]
+    public void UpdateInFindAll()
+    {
+        Test(collection =>
```
pictos, line 10: I'm not sure if this represents a [recursion](https://www.w3schools.com/c/c_functions_recursion.php#:~:text=Recursion%20is%20the%20technique%20of,is%20to%20experiment%20with%20it.), when do you say "recursion" you mean the idea of the method calling it self?

Joy-less: I mean, previously the functions opened the database, did stuff and closed it. Like this:
```cs
OpenDatabase();
DoStuff();
CloseDatabase();
```
This means when you have an `Update()` in a `FindAll()`, it did something like this:
```cs
OpenDatabase();
Loop {
    OpenDatabase();
    DoStuff();
    CloseDatabase();
}
CloseDatabase();
```
Which sounds like recursion to me, I don't know if I mislabelled it.

pictos: Oh, I know that as an iterator. But just names, thanks for clarifying it to me (:

pictos (approved):
Thanks for this PR

PR #2553, opened 2024-10-11
ReaderWriterLockSlim to lock

milo-tcghl:


milo-tcghl:
> please justify this change

When using LiteDb, we write more data than we read. Because LiteDb uses ReaderWriterLockSlim, the write lock will block the read lock, causing the read lock to time out. Therefore, we need to replace ReaderWriterLockSlim with a lock.

milo-tcghl:
> i'm not sure i understand. write locks will continue to block reads, but reads will block reads too, and now there will be no more timeout, which i would regard as a fairly important db reliability mechanism. this doesn't seem very productive.

You are right. If there are performance problems with lock, please give us a good technical solution. Thank you

milo-tcghl:
> > i'm not sure i understand. write locks will continue to block reads, but reads will block reads too, and now there will be no more timeout, which i would regard as a fairly important db reliability mechanism. this doesn't seem very productive.
> 
> You are right. If there are performance problems with lock, please give us a good technical solution. Thank you

How can we resolve the timeout problem?

milo-tcghl:
Can we use the lock strategy int the latest litedb's version?

PR #2557, opened 2024-10-22
Replace Mutex name hashing with URI escaping (Fix #2543)

### LINKED ISSUE #2543: [BUG]
LiteDB version 5.0.21
Microsoft Windows Embedded Standard  6.1.7601.NULL, 64 bit.

We've developed an application in .Net Core 6 that works fine on most machines, but I got a report that it fails with this exception on certain machines:

```
Internal.Cryptography.CryptoThrowHelper+WindowsCryptographicException: Unknown error (0xc1000001)
at Interop.BCrypt.BCryptAlgorithmCache.GetCachedBCryptAlgorithmHandle(String hashAlgorithmId, BCryptOpenAlgorithmProviderFlags flags, Int32& hashSizeInBytes)
at Internal.Cryptography.HashProviderCng..ctor(String hashAlgId, ReadOnlySpan`1 key, Boolean isHmac)
at System.Security.Cryptography.SHA1.Implementation..ctor()
at LiteDB.StringExtensions.Sha1(String value)
at LiteDB.SharedEngine..ctor(EngineSettings settings)
at LiteDB.ConnectionString.CreateEngine(Action`1 engineSettingsAction)
at LiteDB.LiteDatabase..ctor(ConnectionString connectionString, BsonMapper mapper)
at LiteDB.LiteDatabase..ctor(String connectionString, BsonMapper mapper)
```

I'm starting LiteDb in shared mode, and as the exception shows, something goes wrong when creating the name for the mutex. Any idea what could be causing this?

The error only happens once in a while, and if the app is restarted it typically works fine again

Joy-less:
`Shared` mode uses a named Mutex to avoid multiple processes reading/writing at the same time. LiteDB names the Mutex to contain the absolute file path of the database. But Mutex names have to be valid file paths, so this would cause an invalid file path like:
```
Global\c:\users\user\documents\github\litedb\litedb.tests\bin\debug\net8\demo.db.Mutex
```

So instead, currently it hashes the file path using SHA1 to turn it into the following:
```
Global\0a1ab52c4b3f3d01730f61b59e7891bc029ab372.Mutex
```

However, this is a hacky solution in my opinion. Someone experienced exceptions (#2543) caused when creating SHA1. This PR avoids the need to hash the path by using URL escaping:
```
Global\c%3A%5Cusers%5Cuser%5Cdocuments%5Cgithub%5Clitedb%5Clitedb.tests%5Cbin%5Cdebug%5Cnet8%5Cdemo.db.Mutex
```

Also, I changed it to use `.ToLowerInvariant()` instead of `.ToLower()` because Windows file paths are case insensitive even for non-ASCII characters (e.g. `É` and `é`).

This PR should fix #2543 since it removes hashing.

### DIFF
--- a/LiteDB/Client/Shared/SharedEngine.cs
+++ b/LiteDB/Client/Shared/SharedEngine.cs
@@ -21,7 +21,7 @@ namespace LiteDB
         {
             _settings = settings;
 
-            var name = Path.GetFullPath(settings.Filename).ToLower().Sha1();
+            string name = Uri.EscapeDataString(Path.GetFullPath(settings.Filename).ToLowerInvariant());
 
             try
             {
--- a/LiteDB/Utils/Extensions/StringExtensions.cs
+++ b/LiteDB/Utils/Extensions/StringExtensions.cs
@@ -37,24 +37,6 @@ namespace LiteDB
             return true;
         }
 
-        public static string Sha1(this string value)
-        {
-            var data = Encoding.UTF8.GetBytes(value);
-
-            using (var sha = SHA1.Create())
-            {
-                var hashData = sha.ComputeHash(data);
-                var hash = new StringBuilder();
-
-                foreach (var b in hashData)
-                {
-                    hash.Append(b.ToString("X2"));
-                }
-
-                return hash.ToString();
-            }
-        }
-
         /// <summary>
         /// Implement SqlLike in C# string - based on
         /// https://stackoverflow.com/a/8583383/3286260


Joy-less:
I should just mention that since this changes the name of the Mutex, shared mode won't lock correctly if you have two processes where one is using the current or older version of LiteDB and one is using a future LiteDB version.

JKamsker:
Super hard to accept prs that are hard to test :/

JKamsker:
Tbh i am unsure about this change. can we have a setting at userlevel that allows to set a flag for this? (Default this behaviour)

JKamsker:
Hey, just following up on this. Would you be able to add a feature flag in the settings to allow reverting to the old behavior if needed? I believe this would address concerns around backward compatibility and make it easier to adopt the change. Let me know your thoughts!

Joy-less:
> Hey, just following up on this. Would you be able to add a feature flag in the settings to allow reverting to the old behavior if needed? I believe this would address concerns around backward compatibility and make it easier to adopt the change. Let me know your thoughts!

This would be a little frustrating, since you'd have to bring back the `Sha1` function again. I don't think it's a big backwards compatibility break, because it's only there to avoid race conditions in shared mode (which is not a commonly used mode) and only when multiple processes are accessing the same database at once (and one of those processes is using an old LiteDB version). As long as LiteDB.Studio is also updated, I think this should be fine.

Issue #2161, opened 2022-03-11
[BUG] Cannot use custom types in Dictionaries


MathieuDR:
**Version**
Which LiteDB version/OS/.NET framework version are you using. **(REQUIRED)**
* .NET 6
* LiteDB 5.0.11
* PopOS (Ubuntu)

**Describe the bug**
When having a dictionary with a custom key type (for example a strongly typed long) it will be unable to cast/convert into the corrext type, even when you have a mapping function for said custom key.

**Code to Reproduce**

Models
```csharp
public partial struct DiscordUserId {
	public DiscordUserId(ulong value) : this((long)value){ }
	public ulong UlongValue => (ulong)Value;
        public long Value { get; }

        public DiscordUserId(long value)
        {
            Value = value;
        }

        public static readonly DiscordUserId Empty = new DiscordUserId(0);

        public bool Equals(DiscordUserId other) => this.Value.Equals(other.Value);
        public override bool Equals(object obj)
        {
            if (ReferenceEquals(null, obj)) return false;
            return obj is DiscordUserId other && Equals(other);
        }

        public override int GetHashCode() => Value.GetHashCode();

        public override string ToString() => Value.ToString();
        public static bool operator ==(DiscordUserId a, DiscordUserId b) => a.Equals(b);
        public static bool operator !=(DiscordUserId a, DiscordUserId b) => !(a == b);
        public int CompareTo(DiscordUserId other) => Value.CompareTo(other.Value);

        class DiscordUserIdSystemTextJsonConverter : System.Text.Json.Serialization.JsonConverter<DiscordUserId>
        {
            public override DiscordUserId Read(ref System.Text.Json.Utf8JsonReader reader, System.Type typeToConvert, System.Text.Json.JsonSerializerOptions options)
            {
                return new DiscordUserId(reader.GetInt32());
            }

            public override void Write(System.Text.Json.Utf8JsonWriter writer, DiscordUserId value, System.Text.Json.JsonSerializerOptions options)
            {
                writer.WriteNumberValue(value.Value);
            }
        }
}

public class TestModel {
	[BsonId]
	public ObjectId Id { get; set; }
	public DiscordUserId UserId { get; set; }
	public string Name { get; set; }
}

public class TestModelWithDictionary {
	[BsonId]
	public ObjectId Id { get; set; }
	public Dictionary<DiscordUserId, List<TestModel>> Dictionary { get; set; }
}
```

Mapping 
```csharp
[...]

PR #588, opened 2017-04-26
Converting Dictionary keys to string using TypeDescriptor

drcolombo:
This solves #546 as @xied75 suggested based on [http://stackoverflow.com/questions/393731/generic-conversion-function-doesnt-seem-to-work-with-guids](http://stackoverflow.com/questions/393731/generic-conversion-function-doesnt-seem-to-work-with-guids)

PR #2568, opened 2024-11-22
Deserialize dictionary key with TypeDescriptor

### LINKED ISSUE #546: InvalidCastException for DeserializeDictionary when Key is of type Guid
Stack trace : 
```
System.InvalidCastException: Invalid cast from 'System.String' to 'System.Guid'.
   at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   at LiteDB.BsonMapper.DeserializeDictionary(Type K, Type T, IDictionary dict, BsonDocument value)
   at LiteDB.BsonMapper.Deserialize(Type type, BsonValue value)
   at LiteDB.BsonMapper.DeserializeObject(Type type, Object obj, BsonDocument value)
   at LiteDB.BsonMapper.Deserialize(Type type, BsonValue value)
   at LiteDB.BsonMapper.ToObject[T](BsonDocument doc)
   at LiteDB.LiteCollection`1.<Find>d__17.MoveNext()
```
Data Type : 
```
    public class QuizData
    {
        public QuizData()
        { }
        public string RollNumber { get; set; }
        public Dictionary<Guid,Choice> QuestionChoiceList { get; set; }
    }
```
Method Call : 
```
_db.GetCollection<QuizData>("collectionName").FindAll();
```
[I think it's throwing because Generic conversion doesn't support GUIDs.](http://stackoverflow.com/questions/393731/generic-conversion-function-doesnt-seem-to-work-with-guids)
Platform : ASP.NET Core 1.1.0 on .NET Core 1.1.0 using .NET Core SDK `1.0.0-preview2-1-003177`
Edit : I got it working by forking and using a very quick-and-dirty fix in `DeserializeDictionary()`
```
Guid temp;
var k = Guid.TryParse(key, out temp) ? 
temp : K.GetTypeInfo().IsEnum ? 
Enum.Parse(K, key) : Convert.ChangeType(key, K);
```
So basically right now you can't have Guid as a TKey in an IDictionary

Joy-less:
`BsonDocument` only supports `string` keys. Since `Dictionary` is serialized as `BsonDocument`, certain types don't work as keys.

This pull request converts the keys to the correct type when deserializing, using the method in #588.

For example, the key `d396d0e8-2763-4ad6-a1c0-f19f254576ec` can now be deserialized as a `Guid`.

This isn't really ideal but it does the job.

Fixes #546 and #2161. Replaces #588.

### DIFF
--- a/LiteDB/Client/Mapper/BsonMapper.Serialize.cs
+++ b/LiteDB/Client/Mapper/BsonMapper.Serialize.cs
@@ -165,7 +165,9 @@ namespace LiteDB
 
             foreach (DictionaryEntry entry in dict)
             {
-                string stringKey = entry.Key as string ?? JsonSerializer.Serialize(Serialize(entry.Key));
+                string stringKey = entry.Key as string
+                    // Serialize key as JSON to support any key type
+                    ?? JsonSerializer.Serialize(Serialize(entry.Key));
 
                 BsonValue bsonValue = Serialize(valueType, entry.Value, depth);
 
--- a/LiteDB/Client/Mapper/BsonMapper.Serialize.cs
+++ b/LiteDB/Client/Mapper/BsonMapper.Serialize.cs
@@ -1,9 +1,7 @@
 ﻿using System;
 using System.Collections;
-using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using static LiteDB.Constants;
 
 namespace LiteDB
 {
@@ -110,7 +108,7 @@ namespace LiteDB
             }
             else if (obj is Enum)
             {
-                if (this.EnumAsInteger)
+                if (EnumAsInteger)
                 {
                     return new BsonValue((int)obj);
                 }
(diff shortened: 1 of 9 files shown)


JKamsker:
Hi @Joy-less 

Great job on the changes! Could you also implement and ensure the following works as expected?

```csharp
public class DataContainer
{
    public Dictionary<(int, string), string> DataDict { get; set; }
}
```

Additionally, as mentioned earlier on Discord, reviewing a PR becomes challenging when it includes unrelated changes (e.g., refactor-renames). For future submissions, keeping PRs focused on a single purpose would help simplify the review process.

Thank you for your efforts on this!

Joy-less:
> Would it be possible to also implement and ensure the following works as expected?
> 
> ```cs
> public class DataContainer
> {
>     public Dictionary<(int, string), string> DataDict { get; set; }
> }
> ```

This does not work as expected. The key should be serialized as JSON for something complex like this. However, that would break compatibility, so it might be a good idea to assume it's a string key if it fails to deserialize as JSON.

Joy-less:
I've changed the pull request to serialize the key as JSON if it's not a string.
One of the tests was failing because of how LiteDB truncates DateTimes to the nearest millisecond. To keep `JsonSerializer` consistent with `BufferWriter`, I changed the test so now it accepts the truncated value as well.
All tests passed now.

JKamsker:
@Joy-less 
Thanks for the changes! 
Can you think of an scenario where the deserialization doesnt work on databases created with a prior (released) version of litedb?

LiteDB/Client/Mapper/BsonMapper.Serialize.cs
```diff
@@ -128,52 +126,55 @@ internal BsonValue Serialize(Type type, object obj, int depth)
...
+            BsonDocument bsonDocument = [];
 
-            foreach (var key in dict.Keys)
+            foreach (DictionaryEntry entry in dict)
             {
-                var value = dict[key];
-                var skey = key.ToString();
+                string stringKey = entry.Key as string
+                    // Serialize key as JSON to support any key type
+                    ?? JsonSerializer.Serialize(Serialize(entry.Key));
```
JKamsker, line 170:
```csharp
TupleDict = new()
  {
      [(2, "xxx")] = "test",
  },
  ```
from the tests produces
```csharp
System.ValueTuple`2[[System.Int32, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib
```

JKamsker, line 170:
```csharp
TupleDict = new()
  {
      [(2, "xxx")] = "test",
  },
  ```
from the tests produces
```csharp
System.ValueTuple`2[[System.Int32, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.String, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib
```

JKamsker (changes requested):

PR #2569, opened 2024-12-01
Fix potential concurrency issue in GetEntityMapper caused by #2566

JKamsker:


JKamsker:
@vonzshik Can you review this?

Joy-less:
Wouldn't it make more sense to use a ConcurrentDictionary here?

JKamsker:
> Wouldn't it make more sense to use a ConcurrentDictionary here?

[It is?](https://github.com/litedb-org/LiteDB/pull/2569/files#diff-14252f56a26e8a0cc5fc29e38f83a062281baff30f8ea745c6289a05aeefe97dR15)

Joy-less:
> > Wouldn't it make more sense to use a ConcurrentDictionary here?
> 
> [It is?](https://github.com/litedb-org/LiteDB/pull/2569/files#diff-14252f56a26e8a0cc5fc29e38f83a062281baff30f8ea745c6289a05aeefe97dR15)

Sorry, I missed that. If it's a ConcurrentDictionary, then couldn't it use `GetOrAdd` to avoid race conditions?

LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs
```diff
@@ -28,10 +28,32 @@ internal EntityMapper GetEntityMapper(Type type)
...
+                throw new LiteException(LiteException.MAPPING_ERROR, $"Error in '{type.Name}' mapping: {e.Message}", e);
+            }
+        }
+        else
+        {
+            if (!_entities.TryGetValue(type, out mapper))
+            {
+                throw new LiteException(LiteException.MAPPER_NOT_FOUND, $"EntityMapper for type {type} was not found.");
+            }
         }
```
vonzshik, line 56: I would have probably done it like that:

```csharp
mapper = new EntityMapper(type, cts.Token);
EntityMapper addedMapper = _entities.GetOrAdd(type, mapper);
if (ReferenceEquals(addedMapper, mapper))
{
	try
	{
		this.BuildEntityMapper(mapper);
	}
	catch (Exception ex)
	{
		_entities.TryRemove(type, out _);
		throw new LiteException("", ex);
	}
}
```

LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs
```diff
@@ -28,10 +28,32 @@ internal EntityMapper GetEntityMapper(Type type)
...
-        cts.Cancel();
```
vonzshik, line 33: I thought you have to cancel cts as otherwise all `WaitForInitialization`  calls will never complete.

LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs
Joy-less: `GetOrAdd` should have an overload that lets you pass a function rather than an object. If you use that, you can create and set up the mapper in the function rather than having the check.

vonzshik: Then there is no way to determine whether `BuildEntityMapper` should be called or not (and it should be only called once per type), since we do not know whether it's a new mapper or not.

vonzshik: `GetOrAdd` isn't thread safe in a sense that it can be called concurrently from multiple threads.

vonzshik: From the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-8.0):

> If you call [GetOrAdd](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-8.0) simultaneously on different threads, valueFactory may be called multiple times, but only one key/value pair will be added to the dictionary.

vonzshik: > It is, that's why ConcurrentDictionary exists.

I meant that that delegate can be called concurrently, so you might easily get to the point where you have multiple threads calling `BuildEntityMapper` for the exact same type but for different mappers, which is probably not a good thing.

JKamsker: In addition to that, ``GetOrAdd`` will only add the mapper when its initialized. But we need the entity to be present during initialization because it may be used in another mapper which requires our first mapper recursively.

Joy-less: > From the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-8.0):
> 
> > If you call [GetOrAdd](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-8.0) simultaneously on different threads, valueFactory may be called multiple times, but only one key/value pair will be added to the dictionary.

Apologies @vonzshik, I didn't know about this. Thanks for bringing this up.
However, looking at the code in `BuildEntityMapper`, it doesn't look like it would be too big of an issue to build the same type twice since it only affects the passed mapper (with the exception of [the event invocation](https://github.com/litedb-org/LiteDB/blob/80d2f2324c6d5dcfd0a0ac3b0da26d55f7665fee/LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs#L139))

vonzshik: > However, looking at the code in BuildEntityMapper, it doesn't look like it would be too big of an issue to build the same type twice since it only affects the passed mapper (with the exception of [the event invocation](https://github.com/litedb-org/LiteDB/blob/80d2f2324c6d5dcfd0a0ac3b0da26d55f7665fee/LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs#L139))

I was mostly going off the previous implementation/behavior, where it was explicitly made so only one thread can build a particular type (before that was enforced by a `lock`, now it's done by waiting for `cts` to be cancelled). Overall it's not that I'm completely against that, it's just I'm really unfamiliar with the code and whether this change can break anything else or just tank the perf. That's the main reason why I argued to just change that dictionary to concurrent dictionary and leave the `lock` as is, since at the very least that code is relatively stable and simple for understanding.

Joy-less (approved):

LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs
```diff
@@ -25,13 +25,29 @@ internal EntityMapper GetEntityMapper(Type type)
...
+        }
+        finally
+        {
+            // Allow the Mapper to be used for de-/serialization
+            cts.Cancel();
         }
-        cts.Cancel();
-        cts.Dispose();
 
         return mapper;
```
vonzshik, line 52: You're still returning here an unbuilt mapper here, you know...

PR #2572, opened 2024-12-02
 Add support for ValueTypes that consist only of Fields #2571

JKamsker:


JKamsker:
This pull request includes changes to add new tests for issue 2570 and modify the logic for including fields in the `GetTypeMembers` method. The most important changes are detailed below:

### New Tests for Issue 2570:

* [`LiteDB.Tests/Issues/Issue2570_Tests.cs`](diffhunk://#diff-d13ebae86a0d76bd9fc4f67482080a437b18901a52022acb25d89c89a3bd9a2cR1-R63): Added new test cases to verify the handling of tuples and structs in the `Person` and `PersonWithStruct` classes. These tests ensure that the database correctly handles these data structures.

### Logic Modification for Including Fields:

* [`LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs`](diffhunk://#diff-14252f56a26e8a0cc5fc29e38f83a062281baff30f8ea745c6289a05aeefe97dL156-R159): Modified the `GetTypeMembers` method to include fields when the type is a value type and has no properties. This change ensures that value types without properties are still correctly mapped.

Joy-less:
From what I can tell, this is just because `BsonMapper` sets `IncludeFields` to `false` by default. So the behaviour is functioning correctly. It would make more sense imo to set `IncludeFields` to `true` by default or to put a warning that `ValueTuple` uses fields not properties.

JKamsker:
Hi Joy-Less,

Thank you for your feedback! The goal of this PR is to make serialization **more intuitive** while preserving current defaults. Classes typically rely on **properties** to define their data, so the default behavior (`IncludeFields = false`) makes sense to avoid unintentionally serializing fields. Structs, on the other hand, often use **fields** for their data, and this change ensures field-only structs like `ValueTuple` are handled correctly. For structs with properties, the logic still respects properties as the primary data source.

Changing `IncludeFields` to `true` by default could break existing applications, so this PR carefully enhances support for **specific cases** like field-only value types without altering established behavior. I believe this strikes a good balance between stability and usability.

Looking forward to your thoughts!  

Best,  
Jonas

Joy-less:
I understand. Please write a comment in the source code clearly explaining this reasoning, then I will approve.

Issue #2525, opened 2024-07-27
[BUG] Detected loop in FindAll({0})


kamilwarowny:
**Version**
.Net 8.0 
LiteDB.dll 5.0.21.0

**Describe the bug**
About 16000 records in file. File size 33MB

**Code to Reproduce**

```csharp
using (var db = new LiteDatabase(dbFilePAth))
{                   
    var col = db.GetCollection<UserHash>(CollectionName);

    var userHashes = col.Query().ToList();

    db.Checkpoint();

    db.Rebuild();

    return userHashes;
}
```

 line  db.Rebuild(); throws error below:

 at LiteDB.Engine.IndexService.FindAll(CollectionIndex index, Int32 order)+MoveNext()  
 at LiteDB.LinqExtensions.<>c__DisplayClass2_0`2.<<DistinctBy>g___|0>d.MoveNext()  
 at LiteDB.Engine.LiteEngine.<>c__DisplayClass5_0.<EnsureIndex>b__0(TransactionService transaction)  
 at LiteDB.Engine.LiteEngine.AutoTransaction[T](Func`2 fn)  
 at LiteDB.Engine.LiteEngine.EnsureIndex(String collection, String name, BsonExpression expression, Boolean unique)  
 at LiteDB.Engine.LiteEngine.RebuildContent(IFileReader reader)  
 at LiteDB.Engine.RebuildService.Rebuild(RebuildOptions options)  
 at LiteDB.Engine.LiteEngine.Rebuild(RebuildOptions options)  
 at LiteDB.LiteDatabase.Rebuild(RebuildOptions options)

holychaos:
I had this problem too. I tried to move the 'db' file out of the original path, run the program once to generate a new 'db' file, and finally overwrite the moved 'db' file back, and everything worked.

marss72:
Just BTW, not regarding your issue, but it might be possible to improve the provided code by maybe better performance and clarity.

If it's really necessary to use List<T>, ignore this one. But you can improve memory efficiency and lazily load all the entries. I am not an expert on LiteDB so this might cause issues since you are rebuilding it right after.

Also, you can modify `col.Query().ToList();` to use `FindAll()` instead of the `Query()`. Try it yourself, you might get a small performance benefit or not, but I haven't measured it.

k0ka:
I have the same problem. It happens if any table on rebuild has more than 2550 records. Here is why:
[...]

Issue #2582, opened 2025-01-18
[BUG] db.Rebuild throws an exception "Detected loop in FindAll({0})"


frabul:
LiteDB version 5.0.20
Windows 11
.net 6.0
 

When calling Rebuild on the database it fails with an exception "LiteDB.LiteException: 'Detected loop in FindAll({0})'" 
To reproduce it try to open the attached db  and Rebuild.
The attached file is 7z zipped file, i hade to add the .gif extension in order to attach it because 7z is forbidden.
  
```
TradesAndOrdersArch = new LiteDatabase(OperationsArchivePath);
TradesAndOrdersArch.Pragma("UTC_DATE", true);
TradesAndOrdersDb = new LiteDatabase(OperationsDbPath);
TradesAndOrdersDb.Rebuild();
```

![Image](https://github.com/user-attachments/assets/f2f13ee4-bdd6-4d03-b443-d39650160bbc)

patryk9200:
I've published a temporary nuget package variant with fix for this bug:
https://www.nuget.org/packages/LiteDB.Fixed/5.0.22

The source code:
https://github.com/patryk9200/LiteDB

materemias:
@patryk9200 thanks for the fix, I've sent you an email with a sample db which still causes the same issue if you could check ;)

tg2k:
I've been holding on version v5.0.17 waiting for this fix for the past year; Is it indeed still not working per @materemias' message? Also @JKamsker is this an issue in the pre-release v6 versions?

JKamsker:
Consolidated into #2827, which combines this report with the related ones and tracks the fix. Closing this as a duplicate — please follow #2827, and comment there if your case is different.

PR #2603, opened 2025-04-25
Fixed bug Detected loop in FindAll({0}) #2525 #2582

patryk9200:
This PR fixes bug when using Rebuild() to rebuild large database with more than 2550 entries in it it fails thinking that it got an infinite loop.

Fixes:
#2525 #2582

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -224,3 +224,5 @@ FakesAssemblies/
 # Uncomment if necessary however generally it will be regenerated when needed
 #!**/packages/repositories.config
 /LiteDB.BadJsonTest
+/LiteDB/Properties/PublishProfiles/FolderProfile.pubxml
+/LiteDB/README.md
--- a/LiteDB.Tests/Issues/Issue2504_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2504_Tests.cs
@@ -1,0 +1,112 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+
+using Xunit;
+
+namespace LiteDB.Tests.Issues
+{
+    public class Person
+    {
+        [BsonId]
+        public int Id { get; set; }
+        public string Name { get; set; }
+
+        public Person(int id, string name)
+        {
+            Id = id;
+            Name = name;
+        }
+    }
+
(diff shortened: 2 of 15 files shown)


JKamsker:
@codex review

grzegorzszymanskiic:
@JKamsker @patryk9200 can this PR be merged ASAP? We are finding many error reports caused by this issue.

JKamsker:
Thanks for working on this. Unfortunately it doesn't fix #2827 yet. We built this branch and ran the reproduction: 3,000 and 16,000 documents with a secondary index, plus the database attached to #2582. All three still fail on this branch; the error just changes from `Detected loop in FindAll({0})` to `the snapshot is disposed`.

The reason: the item-count bound derived from the source only reaches the insert path. `EnsureIndex` during the rebuild still uses the temporary engine's limit (`Index.cs:32-33`). The new catch in `AutoTransaction` then rolls back the rebuild's own transaction and closes the engine, so the fallback runs on a disposed snapshot. The tests use three documents, which stays below the 2,550 bound, so they don't hit this.

There are also changes that can't ship as they are:
- the package is renamed to `LiteDB.fixed`;
- every open deletes `<name>-temp*.db` files in the data folder, which also matches unrelated files such as `orders-template.db`;
- old backups are pruned silently;
- auto-rebuild is triggered by matching exception text.

The fix direction is described in #2827: make the bound account for pages that are allocated but not yet flushed, or pass a source-derived bound into the `EnsureIndex` path during the rebuild, and add a 3,000-document regression test with a secondary index. A focused PR along those lines against `dev` would be very welcome.

JKamsker:
Closing this implementation without merging. The underlying rebuild / FindAll loop-bound issue remains open and tracked in #2827, which consolidates #2525 and #2582. A tracking update there links this PR and preserves the intended follow-up.

The current change includes broad automatic recovery and file cleanup that cannot ship as written; cleanup can match files outside those owned by the rebuild. A focused replacement for the loop-bound problem, with regression coverage for more than 2,550 documents plus secondary indexes and validation of the vector rebuild path, is welcome. Please preserve and credit the useful regression tests and original contribution from this PR.

PR #2607, opened 2025-05-20
Added missing modulo operator to BsonValue

coldays:
BsonExpressionOperators assumes that you can use the modulo operator % on two BsonValue objects as seen in BsonExpressionOperators.cs:98:
```c#
/// <summary>
/// Mod two number values
/// </summary>
public static BsonValue MOD(BsonValue left, BsonValue right)
{
    if (left.IsNumber && right.IsNumber)
    {
        return left % right;
    }

    return BsonValue.Null;
}
``` 
This pull request implements this operator overload similar to the other arithmetic operator overloads.

Without this change this expression `Select 1.0 % 2.0;` would result in `Specified cast is not valid.`

### DIFF
--- a/LiteDB/Document/BsonValue.cs
+++ b/LiteDB/Document/BsonValue.cs
@@ -500,6 +500,17 @@ namespace LiteDB
             return left.AsDouble / right.AsDouble;
         }
 
+        // %
+        public static BsonValue operator %(BsonValue left, BsonValue right)
+        {
+            if (!left.IsNumber || !right.IsNumber) return BsonValue.Null;
+            if (left.IsInt32 && right.IsInt32) return left.AsInt32 % right.AsInt32;
+            if (left.IsInt64 && right.IsInt64) return left.AsInt64 % right.AsInt64;
+            if (left.IsDecimal && right.IsDecimal) return left.AsDecimal % right.AsDecimal;
+
+            return left.AsDouble % right.AsDouble;
+        }
+
         public override string ToString()
         {
             return JsonSerializer.Serialize(this);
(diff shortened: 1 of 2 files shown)


coldays:
Added to the `Expressions_Scalar_Operator `unit test. The modulo operator was working for int values, probably because it tries using the int implicit operator, but failed for double or decimal.
The lines added to the unit test would fail before overloading the modulo operator for `BsonValue`, but now it runs successfully.

LiteDB/Document/BsonValue.cs
```diff
@@ -500,6 +500,17 @@ public static implicit operator BsonValue(DateTime value)
...
 
+        // %
+        public static BsonValue operator %(BsonValue left, BsonValue right)
+        {
+            if (!left.IsNumber || !right.IsNumber) return BsonValue.Null;
+            if (left.IsInt32 && right.IsInt32) return left.AsInt32 % right.AsInt32;
+            if (left.IsInt64 && right.IsInt64) return left.AsInt64 % right.AsInt64;
+            if (left.IsDecimal && right.IsDecimal) return left.AsDecimal % right.AsDecimal;
+
+            return left.AsDouble % right.AsDouble;
```
JKamsker, line 511: What happens if one of both are a string?

coldays: Then it will return BsonValue.Null, since the first if check will return false on IsNumber from either left or right

coldays:
@JKamsker is there anything else you would like me to change for this PR to be ok?

JKamsker (approved):

JKamsker:
Looks good, thanks for your help!

Issue #1759, opened 2020-07-20
[BUG] BufferExtensions.ToBytes() methods crash on Android ARMv7 with Unity IL2CPP


TigerHix:
**Version**
v5.0.8

**Describe the bug**
Unity app with LiteDB assembly compiled to C++ with Unity's [IL2CPP](https://docs.unity3d.com/Manual/IL2CPP.html) would crash with `signal 7 (SIGBUS), code 1 (BUS_ADRALN)` (Memory misalignment) on Android's ARMv7 ABI.

This bug is due to the memory alignment requirement of ARM gets violated by methods in `BufferExtensions`, for instance [`ToBytes(this Int64 value, byte[] array, int startIndex`](https://github.com/lbnascimento/LiteDB/blob/master/LiteDB/Utils/Extensions/BufferExtensions.cs#L118):

```
        /// <summary>
        /// Copy Int64 bytes direct into buffer
        /// </summary>
        public static unsafe void ToBytes(this Int64 value, byte[] array, int startIndex)
        {
            fixed (byte* ptr = &array[startIndex])
            {
                *(Int64*)ptr = value;
            }
        }
```

This would cause a memory misalignment on 32-bit architecture. Similarly for `ToBytes(this UInt64 value, byte[] array, int startIndex` and `ToBytes(this Double value, byte[] array, int startIndex` as well.

One workaround is to simply rewrite the methods in a safe way:

```
        /// <summary>
        /// Copy Int64 bytes direct into buffer
        /// </summary>
        public static void ToBytes(this Int64 value, byte[] array, int startIndex)
        {
            var bytes = BitConverter.GetBytes(value);
            for (var i = 0; i < bytes.Length; i++)
            {
                array[startIndex + i] = bytes[i];
            }
        }
```

But this would obviously degrade the performance.

It should be noted that the crash only happens when Unity IL2CPP's [C++ Compiler Configuration](https://docs.unity3d.com/ScriptReference/Il2CppCompilerConfiguration.html) is set to `Release`, which probably makes sense as the release configuration may have disallowed unaligned access using a flag like GCC's [`no-unaligned-access`](http://www.keil.com/support/man/docs/armclang_ref/armclang_ref_sam1444138667173.htm).

**Code to Reproduce**
```
public class Test : MonoBehaviour
{
    
    void Start()
    {
        var db = new LiteDatabase(
            new ConnectionString
            {
                Filename = Path.Combine(Application.persistentDataPath, "test.db")
            }
[...]

PR #2696, opened 2025-10-01
Use stack spans for serializer numbers

JKamsker:
## Summary
- switch numeric deserialization in `BufferReader` to use stack-allocated spans with `BinaryPrimitives` and keep pooled buffers only for large reads
- mirror the stack span write path in `BufferWriter` so small numeric writes avoid the array pool while still handling multi-segment copies
- add a regression test that writes and reads numbers across multiple slices to exercise the new stack paths

## Testing
- dotnet test LiteDB.Tests -f net8.0

------
https://chatgpt.com/codex/tasks/task_e_68cfd6f1ea80832a91c26680b1edb837

## Summary by CodeRabbit

- **Improvements**
  - Improved serialization and deserialization of numeric values, including integers, floating-point numbers, and decimals.
  - Enhanced handling of values that span internal buffer boundaries.
  - Reduced memory allocations during numeric serialization operations.

- **Bug Fixes**
  - Improved reliability when reading and writing data across segmented buffers.
  - Strengthened buffer cleanup after failed operations to help prevent resource leaks.

- **Performance**
  - Added benchmarking coverage for numeric serialization and buffer access patterns.


JKamsker (commented):
Merge-readiness update for `fcbae832daa6`.

Changes needed before merge:

- Remove the capturing `SpanFiller` lambdas from numeric writes. Each callback captures the value, introducing closure/delegate allocations on the hot path and defeating the allocation goal of the span rewrite.
- Compare both throughput and allocations against current dev for contiguous and segmented buffers, including integer, floating-point and decimal values. Run production measurements with TestingEnabled=false in separate outputs.
- Keep byte-order, short-buffer and ownership/failure-path coverage across the declared targets. Update against current dev and rerun CI after the change.

The existing focused tests passed. Keep this draft until the hot path is allocation-free as intended and the performance evidence supports the change.

JKamsker:
Addressed the merge-readiness feedback in `eec5b6ae` (with current `dev` merged in `164d7bef`):

- Removed the generic delegate/capturing-lambda number paths. Dedicated inlined contiguous paths now write directly, while no-inline segmented fallbacks use stack spans.
- Removed per-value decimal allocations on .NET 8+ and retained the compatible fallback for `netstandard2.0`/legacy targets.
- Added byte-order, every-number segment-crossing, short-buffer, stale-ownership, and zero-allocation regression coverage.
- Added a production serializer benchmark covering Int32, Double, and Decimal reads/writes in contiguous and boundary-crossing segmented buffers.

BenchmarkDotNet ShortRun, .NET 10, Release, `TestingEnabled=false`, 256 values/operation, separate outputs; baseline is current `dev` at `a50661a9`:

| Scenario | `dev` | PR |
|---|---:|---:|
| Int32 read contiguous / segmented | 945.5 ns / 9,189.7 ns | 420.6 ns / 4,747.7 ns |
| Int32 write contiguous / segmented | 826.4 ns / 9,085.2 ns | 480.2 ns / 4,877.7 ns |
| Double read contiguous / segmented | 974.4 ns / 8,660.5 ns | 657.3 ns / 4,706.1 ns |
| Double write contiguous / segmented | 1,530.3 ns / 9,935.0 ns | 653.0 ns / 4,735.4 ns |
| Decimal read contiguous / segmented | 5,951.1 ns / 48,099.0 ns | 3,191.9 ns / 20,700.8 ns |
| Decimal write contiguous / segmented | 4,361.7 ns / 45,581.6 ns | 2,910.8 ns / 20,166.6 ns |

The benchmark's fixed reader/writer harness allocation is 56 B contiguous and 88 B segmented. Decimal dropped from 10,296/10,328 B on `dev` to that fixed overhead; the focused regression test independently reports 0 B allocated inside repeated numeric serialization for both layouts.

Validation:
- Focused serializer/ownership/allocation tests: 30 passed
- Full .NET 8 tests: 873 passed, 7 skipped
- Full .NET 10 tests: 873 passed, 7 skipped
- Repro-runner tests: 18 passed
- Release solution build with `TestingEnabled=true`: passed
- Production library build with `TestingEnabled=false` for `netstandard2.0`, `net8.0`, and `net10.0`: passed
- C# changed-file size check and `git diff --check`: passed

The local solution-wide command could not execute net462/net481 because this Linux host has no Mono runtime; those assemblies compiled successfully, and the newly triggered GitHub Windows matrix will exercise them.

JKamsker:
This PR removes the small-number pool path in `BufferReader`/`BufferWriter`, but a current `dev` audit found a few related allocations **outside** those classes that are worth including here if they fit cleanly, or otherwise tracking as a follow-up (#2960):

- `BufferSliceExtensions.Write(float)` uses `BitConverter.GetBytes(value).CopyTo(...)` -> temporary `byte[4]`;
- `BufferSliceExtensions.Write(float[] ...)` repeats that allocation **for every vector element**;
- `BufferSliceExtensions.ReadDecimal` constructs `new int[] { a, b, c, d }`.

These matter especially for vector/index-key serialization because the per-float allocation multiplies by vector dimension.

They should be replaceable with span/native primitive writes (or the existing no-allocation `ToBytes` helpers), while retaining the existing native-endian/alignment compatibility required by #1759.

I'd add focused allocation benchmarks for scalar float, decimal and e.g. 128/768/1536-dimension vector index-key round trips.

PR #2699, opened 2025-10-01
Add pooled buffer serialization overloads

JKamsker:
## Summary
- add `Memory<byte>` and `IBufferWriter<byte>` serialization overloads in `BsonSerializer` to reuse caller-provided buffers without extra allocations
- verify round-tripping via `ArrayBufferWriter<byte>` and pooled buffers with new integration tests
- document the API in the README and add the `System.Memory` dependency needed for netstandard builds

## Testing
- dotnet build LiteDB.sln
- dotnet test LiteDB.Tests/LiteDB.Tests.csproj -f net8.0

------
https://chatgpt.com/codex/tasks/task_e_68cfd6f5cd78832ab3a6d54ea97f0040

## Summary by CodeRabbit

* **New Features**
  * Added BSON serialization options that write directly to caller-provided memory buffers or buffer writers.
  * Serialization now reports the number of bytes written and validates insufficient destinations.

* **Documentation**
  * Added README guidance and examples for the new BSON serialization options.

* **Bug Fixes**
  * Updated the `netstandard2.0` memory package dependency for compatibility.

* **Tests**
  * Added coverage for buffer slicing, insufficient capacity, unsupported memory backing, and failed writes.


JKamsker (commented):
Merge-readiness update for `7e95ea17aa88`.

Changes needed before merge:

- Compute the serialized size once and pass it through an internal write helper. The byte-array and IBufferWriter entry points currently calculate the recursive size and then delegate into a Memory overload that calculates it again.
- Repeat the performance comparison against current dev for both the existing byte-array API and the pooled APIs. The previous local x64 Release .NET 8 production comparison against dev `a1612908` measured 423.1 ms versus 529.0 ms for 5,000 serializations (median of seven warmed-up rounds), and 40,744 versus 46,400 allocated bytes per call, for a document with 100 nested records. This is one microbenchmark, but it demonstrates a regression in the existing API.
- Add coverage for undersized destinations, sliced buffers, non-array-backed memory and failed writes without advancing the writer. Document that the memory must be backed by a managed array and that IBufferWriter is omitted from the netstandard2.0 API.

The basic byte-output tests and CI passed. Keep the PR draft until the extra traversal and remaining validation are addressed.

JKamsker:
Addressed the merge-readiness feedback in `f0c0f99e` (with current `dev` merged in `c632bb3e`):

- compute BSON size once per public entry point and pass the validated slice to the shared write helper
- preserve the direct byte-array fast path
- cover undersized and sliced pooled memory, non-array-backed memory, matching `ArrayBufferWriter` output, and no `Advance` after writer validation failures
- document the managed-array requirement and the `netstandard2.0` omission

Validation:
- `python3 scripts/check-csharp-size.py --base origin/dev`
- `dotnet build LiteDB.sln -c Release --no-restore -p:TestingEnabled=true`
- focused buffer tests on net8.0 and net10.0: 6/6 passed on each
- full net8.0 suite: 870 passed, 7 skipped
- full net10.0 suite: 870 passed, 7 skipped

Performance check: Linux x64, Release .NET 8 production builds, tiered compilation disabled, 1,000 warmups, median of seven 5,000-operation rounds, document containing 100 nested records, compared in the same process against `dev` `a50661a9`:

| API | Median / 5,000 | Allocated / op |
| --- | ---: | ---: |
| dev `byte[]` | 886.7 ms | 58,744 B |
| PR `byte[]` | 878.3 ms | 58,744 B |
| PR `Memory<byte>` | 879.3 ms | 46,720 B |
| PR `IBufferWriter<byte>` | 887.3 ms | 46,720 B |

The earlier byte-array regression is no longer present, and the pooled entry points avoid the output-array allocation.

JKamsker:
Addressed the follow-up CodeRabbit style review in 7a897b12 by grouping System imports before LiteDB imports. Validation: all LiteDB target frameworks build successfully, `git diff --check` passes, and `BsonSerializer_Buffer_Tests` passes 6/6 on .NET 8.

Issue #2670, opened 2025-09-28
[BUG] Regression: Find(x => ids.Contains(x.id)) can throw on .NET 10, breaking LINQ-to-Bson translation


LPEdwin:
**Version**
- LiteDB v5.0.21
- .NET 10

**Summary**
On .NET 10 the common pattern

```csharp
var ids = new[] { 1 };
var result = col.Find(x => ids.Contains(x.Id));
```
can fail because the compiler sometimes chooses the span-based overload. The captured array is implicitly converted to ReadOnlySpan<int> and the expression tree contains an op_Implicit(...) node. LiteDB's LINQ-to-Bson translator does not currently handle that shape and throws when trying to translate the lambda.

**Why this happens**

The C# overload resolution rules changed: arrays can implicitly convert to ReadOnlySpan<T> and the compiler may pick MemoryExtensions.Contains(ReadOnlySpan<T>, T) over Enumerable.Contains(IEnumerable<T>, T).
See: https://learn.microsoft.com/dotnet/core/compatibility/core-libraries/10.0/csharp-overload-resolution

**Workaround**
Convert to an IEnumerable<T> explicitly:

```csharp
var ids = new[] { 1 }.AsEnumerable();
var result = col.Find(x => ids.Contains(x.Id));
```

**Expected behavior**

The same code should translate to the same BsonExpression on .NET 9 and .NET 10 (e.g. @0 ANY = @1).

**Proposed fix**

Add a MemoryExtensionsResolver mapping MemoryExtensions.Contains -> @0 ANY = @1 (optionally also IndexOf, StartsWith, EndsWith).

In the LINQ expression visitor, narrowly unwrap op_Implicit conversions only when the conversion is using types of the following group ReadOnlySpan<T> / Span<T> / ReadOnlyMemory<T> / Memory<T>.
Use reflection checks with hardcoded type name comparisons so no project multitargeting is required.

JKamsker:
Wow, that's a fun one - would you mind making a fix for that + tests?
Would be nice to have this in the upcoming v6 release.
Make sure to target dev

LPEdwin:
Yes I can do that.

JKamsker:
Fixed in https://github.com/litedb-org/LiteDB/pull/2703

PR #2710, opened 2025-10-06
Fix/2670 memoryextensions net10 followup

LPEdwin:
This PR is a follow-up to #2670.

The previous patch on `dev` addressed part of the issue, but missed an additional case.

No new test was added, since `LinqBsonExpression_Tests.Linq_Enumerables` was already failing under .NET 10 and now passes with this fix.

The root cause: `Contains` can be invoked with 3 parameters. When the element type does not implement `IEquatable<T>`, a default comparer (`null`) is passed as the third parameter. This scenario was not handled by the previous patch.

Changes:
- Added .NET 10 target framework to the test project
- Added support for 3-parameter `Contains`

Refs #2670

### DIFF
--- a/.github/workflows/_reusable-ci.yml
+++ b/.github/workflows/_reusable-ci.yml
@@ -8,9 +8,9 @@ on:
         type: string
         default: |
           [
-            { "display": ".NET 8",  "sdk": "8.0.x",           "framework": "net8.0", "includePrerelease": false },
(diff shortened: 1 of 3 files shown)


JKamsker:
Thank you for pointing that out! Somehow that must have slipped through.
Anyway, i took the chance to ensure dotnet 10 CI builds and tests correctly now.

Issue #2523, opened 2024-07-25
Regression [BUG] LiteDB.LiteException: ReadFull must read PAGE_SIZE bytes [{0}]


michelmoorlag:
**Version**
On 5.0.20 everything works fine but when I upgrade to 5.0.21 I can't save anything

**Describe the bug**
When I try this code:
```csharp
var collection = _liteDatabaseAsync.GetCollection<T>();
return collection.UpsertAsync(entity);
```
it will throw this exception: {LiteDB.Async.LiteAsyncException: LiteDB encounter an error. Details in the inner exception. ---> LiteDB.LiteException: ReadFull must read PAGE_SIZE bytes [{0}]

**Code to Reproduce**
I have a repositry that implements this code that used to run fine:
```csharp
public Task SaveAsync<T>(T entity) where T : DataObject
{
    try
    {
        var collection = _liteDatabaseAsync.GetCollection<T>();
        return collection.UpsertAsync(entity);
    }
    catch (Exception ex)
    {
        Debug.WriteLine(ex);
        throw;
    }
}
```
where _liteDatabaseAsync is of type: ILiteDatabaseAsync created like this
`_liteDatabaseAsync = new LiteDatabaseAsync(_connectionString);`

**Expected behavior**
I expect that the new versions are backwards comaptible and that above code inserts or updates the given entity without exception.

**Screenshots/Stacktrace**
```
{LiteDB.Async.LiteAsyncException: LiteDB encounter an error. Details in the inner exception. ---> LiteDB.LiteException: ReadFull must read PAGE_SIZE bytes [{0}]
  at LiteDB.Constants.ENSURE (System.Boolean conditional, System.String format, System.Object[] args) [0x00022] in <74012b958fcc45cfa90d6855b44c81c5>:0 
  at LiteDB.Engine.DiskService+<ReadFull>d__23.MoveNext () [0x000b5] in <74012b958fcc45cfa90d6855b44c81c5>:0 
  at LiteDB.Engine.WalIndexService+<>c__DisplayClass19_0+<<CheckpointInternal>g__source|0>d.MoveNext () [0x000d5] in <74012b958fcc45cfa90d6855b44c81c5>:0 
  at LiteDB.Engine.DiskService.WriteDataDisk (System.Collections.Generic.IEnumerable`1[T] pages) [0x0006e] in <74012b958fcc45cfa90d6855b44c81c5>:0 
[...]

PR #2716, opened 2025-10-25
Fix ReadFull must read PAGE_SIZE bytes [{0}]

isbdnt1:


isbdnt1:
https://github.com/litedb-org/LiteDB/issues/2523

PR #2717, opened 2025-10-25
Fix ReadFull must read PAGE_SIZE bytes [{0}]

isbdnt1:


isbdnt1:
https://github.com/litedb-org/LiteDB/issues/2523

JKamsker:
Can you provide tests which prove that you fix did fix it?

isbdnt1:
> Can you provide tests which prove that you fix did fix it?

I'm not sure how to provide tests. This is an occasional bug. The trigger reason is that failing to flush immediately after WriteStream causes ReadStream to not read the latest Page.

image

isbdnt1:
WriteStream at
image

ReadStream at
image

isbdnt1:
When the bug occurs, bytesRead is zero.
image

JKamsker:
Even tho we can't consistently reproduce the issue out in the wild, i was able to craft a regression test which drips when your fix is not in place. Thank you for your contribution!

PR #2722, opened 2025-11-02
Feat(Mapper): Support [Key] and [NotMapped] attributes from DataAnnotations

mrdevrobot:
This PR enhances the `BsonMapper` to recognize common attributes from `System.ComponentModel.DataAnnotations`, improving interoperability with other frameworks like Entity Framework.

**Changes:**

-   **`[Key]` as `[BsonId]`:** The `[Key]` attribute is now recognized as an alias for `[BsonId(AutoId = false)]`.
    
-   **`[NotMapped]` as `[BsonIgnore]`:** The `[NotMapped]` attribute is now recognized as an alias for `[BsonIgnore]`.
    

To avoid adding new dependencies to older target frameworks (like .NET Standard 2.0), this functionality is conditionally compiled and **only active for projects targeting .NET 8 or newer** (`#if NET8_0_OR_GREATER`).

**Verification:**

-   Added new unit tests in `CustomMapping_Tests.cs` to verify that `[Key]` and `[NotMapped]` attributes are correctly handled.
    
-   Added a new benchmark (`InsertionIgnoreExpressionPropertyComponentModelBenchmark`) to measure insertion performance when using property exclusion attributes.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in support for mapping entities with standard data annotations.
  - Added mapper-specific registration of custom attributes for identifying document IDs and ignoring members.
  - Supports registration by attribute type or fully qualified name.
  - Custom mappings work with inherited and overridden members and remain isolated between mapper instances.

- **Bug Fixes**
  - Ensured explicitly configured ID attributes take precedence over inherited or automatic ID settings.


LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs
```diff
@@ -57,9 +61,15 @@ internal EntityMapper GetEntityMapper(Type type)
...
-        var ignoreAttr = typeof(BsonIgnoreAttribute);
+    {
+#if NET8_0_OR_GREATER
+        var idAttrs = new List<Type> { typeof(BsonIdAttribute), typeof(KeyAttribute) };
+        var ignoreAttrs = new List<Type> { typeof(BsonIgnoreAttribute), typeof(NotMappedAttribute) };
+#else
+        // in netstandard2.0 KeyAttribute and NotMappedAttribute are not available without adding extra dependencies
+        var idAttrs = new List<Type> { typeof(BsonIdAttribute) };
+        var ignoreAttrs = new List<Type> { typeof(BsonIgnoreAttribute) };
+#endif
```
JKamsker: Make them
- Private static fields
- Netstandard: Use reflection to find them maybe?
- (Optional)Configurable from outside - maybe users want to define them (just an idea)

README.md
JKamsker, line 1: Way too much change here - what's going on?

JKamsker:
Thanks for your contribution! Please see my comments.

LiteDB/Client/Mapper/BsonMapper.GetEntityMapper.cs
mrdevrobot: Hi @JKamsker in the original method there are a few local variabiables for managing those and others attributes. Do you want to make all of them static?

Using reflection without type checking means rely on string representation of such attributes. Even if it is possible it doesn't sound as a good design choice.

The configuration of custom attributes is possible and should be possible to ne managed in the BsonMapper.Global.
I'd rather open a separate issue for developments.

JKamsker: > Hi @JKamsker in the original method there are a few local variabiables for managing those and others attributes. Do you want to make all of them static?

If they dont change and are readonly? Absolutely yes!

> Using reflection without type checking means rely on string representation of such attributes. Even if it is possible it doesn't sound as a good design choice.

<``NET 8`` is not a priority in terms of performance. We want feature parity between those builds, even if it costs us performance in old versions. Try to use reflection sparingly tho - cache what you can. And only for sub net8.

README.md
mrdevrobot: thought to add a v6 section with details of the improvements, but some layout fixes from the md editor also came in. I revert to previous version with only the v6

JKamsker:
@codex review

JKamsker (changes requested):
Merge-readiness update for `8ef6f26eb410`.

Changes needed before merge:

- Honor inherited mapping attributes on overridden members. The new `GetCustomAttributes(false)` lookup misses inherited attributes, including the reviewed NotMapped override case. Cover Key and NotMapped on base/derived types and their precedence with LiteDB's existing attributes.
- Make attribute registration and lookup safe under concurrent access. Locking writes while reading mutable shared lists without the same synchronization does not provide a safe publication/enumeration contract; use a consistent lock or immutable snapshots.
- Add focused inheritance and concurrent registration/mapping tests, then update against current dev and run CI. Preserve the newly merged mapper extension points from #1212.

JKamsker:
Addressed the requested changes in 03a806c8:

- Merged current upstream dev (a50661a9) into the contributor branch, preserving the mapper extension points from #1212.
- Made DataAnnotations support opt-in per BsonMapper via UseDataAnnotations(), with Key and NotMapped matched by full name on every target framework.
- Made custom attribute registration per mapper and concurrency-safe with lock-published immutable array snapshots.
- Honored inherited attributes on overridden members and retained BsonId/AutoId precedence.
- Removed unrelated README, Code of Conduct, benchmark, and solution changes from the effective PR diff.
- Added focused opt-in, mapper-isolation, inheritance, precedence, validation, and concurrent registration/mapping coverage.

Validation:
- dotnet restore LiteDB.sln
- dotnet build LiteDB.sln -c Release --no-restore -p:TestingEnabled=true (all target frameworks built)
- focused mapper tests: 9 passed on net8.0 and 9 passed on net10.0
- full LiteDB.Tests: 873 passed, 7 skipped on net8.0; 873 passed, 7 skipped on net10.0
- LiteDB.ReproRunner.Tests: 18 passed on net8.0
- staged C# size check passed

The solution-wide test command also attempted net462/net481 execution, but this Linux runner has no Mono host. Both .NET Framework test assemblies compiled successfully in the Release build.

JKamsker (approved):
All requested changes are addressed. The PR is current with dev, all review threads are resolved, and the full 45-job CI matrix is green.

PR #2745, opened 2026-04-01
Provide ToString implementation of ConnectionString

rklfss:
With this PR, ToString() of ConnectionString outputs the connection string representation as per documentation https://www.litedb.org/docs/connection-string/.

TODO: Documentation says for initial size it is "InitialSize=", but code was "Initial Size=" with space. Either of them is wrong.

## Summary by CodeRabbit

- **Enhancements**
  - Improved connection-string parsing and serialization for filenames containing spaces, quotes, backslashes, and ambiguous characters.
  - Preserves additional settings, including initial size, collation, memory profile, cache size, and transaction limits.
  - Passwords are redacted by default, with explicit password-inclusive serialization available.
  - Preserves empty passwords and quoted values accurately.
  - Culture name matching is now case-insensitive.

- **Bug Fixes**
  - Invalid quoted values, including missing separators and unterminated values, now report a format error.

### DIFF
--- a/LiteDB.Tests/Database/ConnectionString_Tests.cs
+++ b/LiteDB.Tests/Database/ConnectionString_Tests.cs
@@ -26,12 +26,12 @@ namespace LiteDB.Tests.Database
 
             // file with spaces with " and ;
             var full = new ConnectionString(
-                @"filename=""c:\only;file\""d\""emo.db""; 
+                @"filename=""c:\only;file\""d\""e=mo.db""; 
                   password =   ""john-doe "" ;
                   initial size = 10 MB ;
                   readONLY =  TRUE;");
 
-            full.Filename.Should().Be(@"c:\only;file""d""emo.db");
+            full.Filename.Should().Be(@"c:\only;file""d""e=mo.db");
             full.Password.Should().Be("john-doe ");
             full.ReadOnly.Should().BeTrue();
             full.InitialSize.Should().Be(10 * 1024 * 1024);
@@ -46,6 +46,44 @@ namespace LiteDB.Tests.Database
             cn.Filename.Length.Should().Be(49);
             cn.Password.Length.Should().Be(512);
 
+        }
+
+        [Fact]
+        public void ConnectionString_ToString()
+        {
+            var empty = new ConnectionString();
+
+            empty.ToString().Should().BeEmpty();
+
+            var onlyfile = new ConnectionString
+            {
+                Filename = @"c:\only file\demo.db",
+            };
+
+            onlyfile.ToString().Should().Be(@"c:\only file\demo.db");
+
+            // filename with =
+            var fileWithSpecials = new ConnectionString
+            {
+                Filename = @"c:\only file\d""e=mo.db",
+            };
+
+            fileWithSpecials.ToString().Should().Be(@"Filename=""c:\only file\d\""e=mo.db""");
+
+            // file with spaces with " and ;
+            var full = new ConnectionString
+            {
+                Filename = @"c:\only;file""d""emo.db",
+                Password = "john-doe ",
+                ReadOnly = true,
+                InitialSize = 10_485_760,
+            };
+
+            full.ToString().Should().Be(@"Filename=""c:\only;file\""d\""emo.db"";Password=""john-doe "";Initial Size=10485760;ReadOnly=True");
+
+            // ToString/Parse round trips
+            var parsed = new ConnectionString(full.ToString());
+            parsed.Should().BeEquivalentTo(full);
         }
     }
 }
--- a/LiteDB/Client/Structures/ConnectionString.cs
+++ b/LiteDB/Client/Structures/ConnectionString.cs
@@ -2,7 +2,7 @@ using LiteDB.Engine;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using static LiteDB.Constants;
+using System.Text;
 
 namespace LiteDB
 {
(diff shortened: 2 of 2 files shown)


JKamsker:

Reviewed exact head `549661af68588d2fdd4912d9a3e5fd2780752429`. Five issues survived reproduction.

### 1. P1 — a password or path ending in backslash corrupts everything after it

~~~csharp
var source = new ConnectionString
{
    Filename = "old.db",
    Password = "secret\\",
    ReadOnly = true,
    Upgrade = true,
    AutoRebuild = true
};

var reparsed = new ConnectionString(source.ToString());
~~~

The generated closing quote is interpreted as escaped. The result is effectively:

~~~text
Password = secret";ReadOnly=True;Upgrade=True;Auto-Rebuild=True
ReadOnly = false
Upgrade = false
AutoRebuild = false
~~~

An encrypted database using that password fails to reopen with “Invalid password.” A filename ending in backslash can similarly select or create the wrong database.

Code: [ConnectionString.cs lines 161–165 and 232–251](https://github.com/litedb-org/LiteDB/blob/549661af68588d2fdd4912d9a3e5fd2780752429/LiteDB/Client/Structures/ConnectionString.cs#L232-L251).

**Suggested direction:** define a symmetric escaping grammar, escape backslashes as well as quotes, and determine escaped quotes from the parity of preceding backslashes.

### 2. P2/security — ToString now leaks passwords implicitly

~~~csharp
logger.LogInformation("Opening {Connection}", connectionString);
Console.WriteLine(connectionString);
~~~

Before this PR those calls printed the type name. Afterwards they print the plaintext password. This can leak database credentials into logs, telemetry, crash reports, and support bundles.

**Suggested direction:** keep ToString redacted and provide an explicitly named method for secret-bearing serialization.

### 3. P2 — significant filename whitespace is removed

~~~csharp
var source = new ConnectionString
{
    Filename = "/tmp/ database.db "
};

var reparsed = new ConnectionString(source.ToString());
// expected: "/tmp/ database.db "
// actual:   "/tmp/ database.db"
~~~

On Unix those are different legal filenames. The exact-head integration repro inserted into the whitespace-named database, reparsed the string, then opened a new empty database at the trimmed path.

Code: [ConnectionString.cs lines 148–150 and 223–225](https://github.com/litedb-org/LiteDB/blob/549661af68588d2fdd4912d9a3e5fd2780752429/LiteDB/Client/Structures/ConnectionString.cs#L223-L225).

### 4. P2 — Valencian collation does not round-trip

LCID 2051 is emitted by the runtime as:

~~~text
ca-ES-VALENCIA/IgnoreCase
~~~

LiteDB's lookup expects case-sensitive `ca-ES-valencia`, so reparsing throws “Invalid culture name.”

**Suggested direction:** compare culture names with OrdinalIgnoreCase or serialize LiteDB's stable LCID mapping.

### 5. P2 — an empty password emits invalid syntax

~~~csharp
new ConnectionString
{
    Filename = "old.db",
    Password = string.Empty
}.ToString();
~~~

produces:

~~~text
Filename="old.db";Password=
~~~

Reparsing throws IndexOutOfRangeException. Either normalize empty to null and omit it, or emit `Password=""` and harden EOF-after-equals parsing.

There is no database page-format change here, but these bugs can prevent an existing encrypted database from opening, discard safety flags, expose credentials, or silently redirect writes to another file.

JKamsker (changes requested):
Thanks for this, and sorry for the wait. The review on Aug 10 found five problems that still need fixing before this can go in:

1. Backslashes aren't escaped symmetrically, so a value ending in `\` swallows the keys after it.
2. `ToString()` includes the password in plain text. `ToString()` gets called implicitly by logging, debuggers and string interpolation, so please keep it redacted and add an explicitly named method for the form that includes secrets.
3. Significant leading or trailing whitespace in the filename is trimmed.
4. Collation culture names don't round-trip case-insensitively (e.g. `ca-ES-VALENCIA`).
5. An empty password is written out as `Password=`.

Please also rebase once #2772 lands. It adds the `memory profile`, `cache size` and `transaction pages` keys, which `ToString()` has to write for the round trip to work. Longer term, this fits the typed connection-string builder proposed in #2837.

JKamsker:
Quick update: #2772 is merged into dev now. When you rebase, `ToString()` also needs to write its three new keys, `memory profile`, `cache size` and `transaction pages`, so that a connection string survives the round trip. The earlier review points (backslash escaping, redacting the password, filename whitespace, culture names, empty password) still apply.

JKamsker (changes requested):
Merge-readiness update for `51420f8b2134`.

The changes-requested decision still applies after the conflict resolution. Please address the remaining contract before merge:

- Make quoting and backslash escaping symmetric, including values ending in a backslash; cover significant filename whitespace, culture-name casing and empty passwords in the round-trip tests.
- Keep the ordinary diagnostic `ToString()` password-redacted, as requested in the earlier review. Provide an explicitly named opt-in API for a connection string that includes secrets.
- Serialize the current `memory profile`, `cache size` and `transaction pages` settings and preserve their effective values on round trip.
- Coordinate the escaping/option contract with #944 and #2757 and rerun focused tests plus current CI.

The current 20 connection-string tests passed during conflict resolution, but they did not clear these outstanding findings.

rklfss:
Added an overload for the plain text password and added the new properties. Empty passwords are handled as well.

For the rest. Sure, i added the test, if the connection string round trips with ToString and the CTOR, but things mentioned are not due the change. Especially
- significant filename whitespace is removed: Connection String filename gets trimmed in the CTOR
- Valencian collation does not round-trip: Guess the name should be compared case-insensitice in the LCID.GetLCID method

If I change it, we sure will have more rounds of "this breaks this and that"...

JKamsker:
Thanks for the contribution! <3

PR #2747, opened 2026-05-07
XONE-7891 preview

mts-liryk:


JKamsker (changes requested):
Do not use Console.Writeline

PR #2752, opened 2026-06-10
Fix #1986: return null/default for Min and Max on empty collections

### LINKED ISSUE #1986: [BUG]Find Max error: Sequence contains no elements
v5.0.10

int max = _db.GetCollection<CodeGroup>(_Collection_CodeGroups).Max(g => g.ShowIndex); // OK

int max = _db.GetCollection<CodeGroup>(_Collection_CodeGroups).Find(g => g.ParentId == parentId).Max(g => g.ShowIndex); //ERROR
Sequence contains no elements

any way to find max field ?

JKamsker:
Refs #1986.

## Summary
- Updates collection `Min`/`Max` aggregation to avoid throwing `Sequence contains no elements` on empty collections.
- Returns `BsonValue.Null` from expression-based overloads when no value exists.
- Returns `default(T)` from generic overloads for empty collections.

## TDD / verification
- Added Issue1986 regression tests for `Min` and `Max` on empty collections.
- Verified targeted issue tests and the full `net8.0` test project.

## Compatibility note
No on-disk format change. This is a runtime behavior fix for empty aggregate results.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `Min` and `Max` queries on empty collections or unmatched filters so they return null or the appropriate default value instead of throwing errors.
  - Improved handling of stored null values and typed aggregate results, including custom deserialization scenarios.
  - Preserved correct minimum and maximum results for collections containing matching records.

### DIFF
--- a/LiteDB.Tests/Issues/Issue1986_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue1986_Tests.cs
@@ -1,4 +1,5 @@
 using System.IO;
+using System.Linq;
 using Xunit;
 
 namespace LiteDB.Tests.Issues;
@@ -91,5 +92,40 @@ public class Issue1986_Tests
 
         Assert.Equal(9, col.Max(x => x.Value));
         Assert.Equal(2, col.Min(x => x.Value));
+    }
+
+    [Fact]
+    public void Typed_min_should_not_turn_a_stored_BSON_null_into_zero()
+    {
+        using var db = new LiteDatabase(new MemoryStream());
+
+        db.GetCollection("e").Insert(new BsonDocument
+        {
+            ["_id"] = 1,
+            ["Value"] = BsonValue.Null
+        });
+
+        var typed = db.GetCollection<Entity>("e");
+
+        // Entity.Value is a non-nullable int. Returning 0 would invent a value
+        // which was never stored, so this incompatible read must fail visibly.
+        var exception = Record.Exception(() => typed.Min(x => x.Value));
+
+        Assert.NotNull(exception);
+    }
+
+    [Fact]
+    public void Filtered_Max_call_from_issue_1986_should_handle_no_matches()
+    {
+        using var db = new LiteDatabase(new MemoryStream());
+        var col = db.GetCollection<Entity>("e");
+        col.Insert(new Entity { Id = 1, Value = 10 });
+
+        int max = -1;
+        var exception = Record.Exception(() =>
+            max = col.Find(x => x.Id == 999).Max(x => x.Value));
+
+        Assert.Null(exception);
+        Assert.Equal(0, max);
     }
 }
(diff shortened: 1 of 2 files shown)


JKamsker:

Reviewed exact head `de06f51c21be31a10fbbeb8d8e32ec64c66f1a75`. Two issues survived reproduction.

### 1. P1 — a real BSON Null is reported as a made-up CLR default

~~~csharp
var raw = db.GetCollection("raw");
raw.Insert(new BsonDocument { ["_id"] = 1, ["Value"] = BsonValue.Null });
raw.Insert(new BsonDocument { ["_id"] = 2, ["Value"] = 7 });

var typed = db.GetCollection<Entity>("raw");

raw.Min("Value");        // BSON Null
typed.Min(x => x.Value); // PR head returns 0
~~~

The query did find a real document. Its value was Null. The new code treats the deserialized null as though no document existed and returns `default(int)`.

That can invent plausible data which was never stored:

~~~text
int      -> 0
bool     -> false
DateTime -> DateTime.MinValue
~~~

Before this PR the incompatible typed read failed visibly; after upgrade it silently returns a business-significant value.

Code: [Aggregate.cs lines 167–170](https://github.com/litedb-org/LiteDB/blob/de06f51c21be31a10fbbeb8d8e32ec64c66f1a75/LiteDB/Client/Database/Collections/Aggregate.cs#L167-L170) and [209–212](https://github.com/litedb-org/LiteDB/blob/de06f51c21be31a10fbbeb8d8e32ec64c66f1a75/LiteDB/Client/Database/Collections/Aggregate.cs#L209-L212).

**Suggested direction:** keep “did a row exist?” separate from the selected BSON value. Return default only for no row. Preserve null for nullable/reference targets and give a clear error for a present Null mapped to a non-nullable value type.

### 2. P2 — the tests do not execute the call from issue #1986

The issue reports:

~~~csharp
col.Find(g => g.ParentId == parentId)
   .Max(g => g.ShowIndex);
~~~

But `Find` returns `IEnumerable<T>`, so that call invokes LINQ's `Enumerable.Max`, not the patched collection aggregate.

With one stored row and a predicate matching nothing:

~~~text
base -> InvalidOperationException
head -> InvalidOperationException
~~~

The added tests call `col.Max(...)` directly on an empty collection, which is a different API path. Merging this as “Fix #1986” would leave the issue's exact code failing.

Tests: [Issue1986_Tests.cs lines 24–64](https://github.com/litedb-org/LiteDB/blob/de06f51c21be31a10fbbeb8d8e32ec64c66f1a75/LiteDB.Tests/Issues/Issue1986_Tests.cs#L24-L64).

**Suggested direction:** first add the issue's exact filtered-call regression. Then either provide a filtered LiteDB aggregate API or document the normal LINQ empty-sequence pattern and narrow the PR's claim.

No storage bytes are changed, but identical old data would produce different—and potentially silently wrong—typed results after upgrade.

PR #2753, opened 2026-06-10
Fix #2255: serialize dictionary keys invariantly

### LINKED ISSUE #2255: [BUG] Inconsistent database entries depending on current culture
**Version**
LiteDB: 5.0.12, dotNet Standard 2.0

**Describe the bug**
When using a `Dictionary<double, double>`  in  a DB collection the key is serialized with the current culture settings. On my German Windows instance `9.9 => 1.23` becomes `"9,9": 1.23` in json/bson format. When fetching data from the db, the dictionary row becomes `99 => 1.23`. Even worse: When adding data with a different culture a new row is inserted with a different string (result of code below):

|_id|Values|
|---|---|
|1|{"9.9":1.23}|
|2|{"9,9":1.23,"10,1":1.23}|

**Code to Reproduce**
```cs
class DatabaseRow
{
    public int Id {get; set;}
    public Dictionary<double, double> Values {get; set;}
}

class App
{
    public void Write()
    {
        System.Globalization.CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
        var connectionString = new ConnectionString("dbfile.db") { Upgrade = true };
        var db = new LiteDatabase(connectionString);
        var table = db.GetCollection<DatabaseRow>("data");
        var row = new DatabaseRow();
        row.Values = new Dictionary<double, double>();
        row.Values[9.9] = 1.23;
        table.Insert(row);
        System.Globalization.CultureInfo.CurrentCulture = CultureInfo.CreateSpecificCulture("de-DE");
        row.Values[10.10] = 1.23;
        table.Update(row);
    }

```

**Expected behavior**
All values of culture dependent data types in the database should be culture independent. Thus read and write operation will always have the same result regardless of the current thread culture.

JKamsker:
Fixes #2255.

## Summary
- Serializes non-string dictionary keys using invariant culture where appropriate.
- Prevents persisted dictionary field names from changing with `CurrentCulture`.
- Keeps existing string-key dictionary behavior unchanged.

## TDD / verification
- Added regression coverage demonstrating culture-dependent dictionary key corruption before the fix.
- Verified the focused Issue2255 tests and the full `net8.0` test project.

## Compatibility note
No LiteDB file format change. The fix stabilizes the string representation of dictionary keys going forward.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dictionary keys now serialize consistently across regional settings.
  * Date/time keys use a stable format that can be reliably round-tripped.
  * Declared key types are respected when converting dictionary keys.
  * Existing dictionary field names are preserved when data is saved again.

### DIFF
--- a/LiteDB.Tests/Issues/Issue2255_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2255_Tests.cs
@@ -90,6 +90,39 @@ public class Issue2255_Tests
         }
     }
 
+    [TypeConverter(typeof(CollidingKeyConverter))]
+    public sealed class CollidingKey
+    {
+        public CollidingKey(int number)
+        {
+            Number = number;
+        }
+
+        public int Number { get; }
+    }
+
+    public sealed class CollidingKeyConverter : TypeConverter
+    {
+        public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
+        {
+            return destinationType == typeof(string) || base.CanConvertTo(context, destinationType);
+        }
+
+        public override object ConvertTo(
+            ITypeDescriptorContext context,
+            CultureInfo culture,
+            object value,
+            Type destinationType)
+        {
+            if (destinationType == typeof(string) && value is CollidingKey key)
+            {
+                return key.Number == 1 ? "same" : "SAME";
+            }
+
+            return base.ConvertTo(context, culture, value, destinationType);
+        }
+    }
+
     private class Row
     {
         public int Id { get; set; }
@@ -172,6 +205,21 @@ public class Issue2255_Tests
         Assert.Equal(2, document.Count);
         Assert.Equal("first value", document["first:42"].AsString);
         Assert.Equal("second value", document["second:42"].AsString);
+    }
+
+    [Fact]
+    public void Dictionary_key_conversion_should_reject_case_insensitive_collisions()
+    {
+        var values = new Dictionary<CollidingKey, string>
+        {
+            [new CollidingKey(1)] = "first",
+            [new CollidingKey(2)] = "second"
+        };
+
+        var exception = Record.Exception(() => new BsonMapper().Serialize(values));
+
+        Assert.NotNull(exception);
+        Assert.Contains("same", exception.Message, StringComparison.OrdinalIgnoreCase);
     }
 
     [Fact]
--- a/LiteDB.Tests/Issues/Issue2255_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2255_Tests.cs
@@ -1,0 +1,84 @@
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using Xunit;
+
+namespace LiteDB.Tests.Issues;
+
+/// <summary>
+/// #2255 - non-string dictionary keys (e.g. Dictionary&lt;double,double&gt;) are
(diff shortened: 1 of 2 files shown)


JKamsker:

Reviewed exact head `20e37e826eb1c1ef656d8539f0066d1e629a286d`. Two compatibility defects survived reproduction.

### 1. P1 — runtime converters can silently merge different dictionary keys

The writer chooses a converter from each key's runtime type, while the reader chooses the converter for the declared `TKey`.

Simplified example:

~~~text
Dictionary<BaseKey, string>

DerivedKey converter: 42 -> "42"
OtherKey converter:   42 -> "42"

BaseKey reader expects:
  "derived:42"
  "other:42"
~~~

Head writes both values to the same BSON field name:

~~~text
[42] = "first"
[42] = "second"  // silently overwrites first
~~~

Verified result:

~~~text
base stored and round-tripped 2 entries
head stored only [42]="second"
head typed read then threw FormatException
~~~

A base-created database remains readable until a head update. That update can permanently discard one dictionary entry; downgrading cannot recover it.

Code: [BsonMapper.Serialize.cs lines 182–189](https://github.com/litedb-org/LiteDB/blob/20e37e826eb1c1ef656d8539f0066d1e629a286d/LiteDB/Client/Mapper/BsonMapper.Serialize.cs#L182-L189).

**Suggested direction:** serialize using the declared key type—the same protocol used for deserialization—and reject duplicate serialized field names before overwriting.

### 2. P2 — new key spellings break existing paths and indexes

Dictionary keys are persisted as BSON field names. They therefore form part of raw query paths and saved index expressions.

Examples of base versus head spelling:

~~~text
DateTimeOffset:
base: 8/10/2026 3:27:33 PM +02:00
head: 08/10/2026 15:27:33 +02:00

Uri:
base: https://example.com/a b?q=x
head: https://example.com/a%20b?q=x

DateOnly:
base: 1/2/2020
head: 2020-01-02
~~~

Concrete database reproduction:

~~~text
base creates unique index on old DateTimeOffset field path
head reads the row
head update renames the BSON field
saved index still points to old path
old-path matches: 1 -> 0
head inserts a duplicate semantic value successfully
~~~

The unique constraint was bypassed because the index expression could no longer see the renamed field. A base rewrite changes the spelling back, so mixed writers oscillate the schema.

**Suggested direction:** preserve legacy spellings for already-round-trippable types. If normalization is required, make it a versioned migration which rewrites fields and rebuilds dependent indexes atomically.

There is no page-format change, but mapper/query/index schema compatibility is broken, and the first finding can cause irreversible data loss.

JKamsker (commented):
Thanks! The Aug 11 commits fix both review findings: the declared key converter is used now, and duplicate field names throw instead of overwriting. A few things before merging:

1. `DateTimeOffset` keys are still written in the current culture to keep the legacy spelling, so under e.g. `de-DE` they probably still don't round-trip through the invariant read path. Please add a `de-DE` `DateTimeOffset` test, plus either an invariant-read fallback or a note that this type keeps the legacy behaviour.
2. The key-collision error uses `LiteException(0, ...)`. Please give it a real error code.
3. Not blocking: keys of other types (e.g. `DateOnly`, `Uri`) can change their stored spelling compared with data written under a non-invariant culture. That's the intended fix, but it needs a line in the release notes, since existing indexes on such paths may be affected.

PR #2754, opened 2026-06-10
Fix #2376: serialize byte-backed enum arrays as arrays

### LINKED ISSUE #2376: [BUG] LiteDB cant save an enum array of type 'byte' but type 'int' works
**Version**
5.0.17

**Describe the bug**
liteDB cannot save Enum Array of type byte 
gives error: "ArgumentException: Object must be an array of primitives." on "collection.Insert();

**Code to Reproduce**
```
 public class MyClass
{
    public MyClass(enumTest enumTest)
    {
        _enumTest =new []{enumTest,enumTest};
    }

    [BsonId]
    public int _id
    {
        get;
        set;
    }
    public enumTest[] _enumTest
    {
        get;
        set;
    }
    public enum enumTest:byte //  <----   Changing this to 'int' fixes issue!
   {
       me,you,us
   } 
}
```

```
 using(var db = new LiteDatabase(dbFilePath))
                {
                    // Get customer collection or create
                    var collection = db.GetCollection<MyClass>("test");
                    var mc = new MyClass(MyClass.enumTest.me);
                    // Insert new customer document (Id will be auto-incremented)
                    collection.Insert(mc);
                    // false means not found! (did not insert)
                b =collection.Update(mc);
                } 
```

**Expected behavior**
give no error and write the enum(byte) array to file

**Screenshots/Stacktrace**
-

**Additional context**
-

JKamsker:
Fixes #2376.

## Summary
- Prevents byte-backed enum arrays from being mistaken for raw `byte[]` binary data.
- Serializes byte-backed enum arrays as BSON arrays, matching other enum array backing types.
- Keeps actual `byte[]` values serialized as binary.
- Preserves legacy binary-stored enum array reads.

## TDD / verification
- Added Issue2376 tests for raw mapper output, typed round-trip, and array-query behavior.
- Added backward-compat coverage proving legacy binary-stored byte enum arrays still deserialize.
- Verified targeted issue tests and the full `net8.0` test project.

## Compatibility note
New writes of byte-backed enum arrays now use the correct BSON array shape. Existing buggy binary-stored values remain readable, but historical data must be re-saved if callers need array-element querying/indexing over old documents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved serialization of byte-backed enum arrays so they are stored and restored as BSON arrays while preserving regular byte-array binary behavior.
  * Ensured dictionary keys serialize consistently across cultures, including database round-tripping and resaving existing data.
  * Added validation to prevent distinct dictionary keys from producing conflicting field names.
  * Preserved compatibility with legacy binary data, non-generic collections, and existing date/time formatting.

### DIFF
--- a/LiteDB.Tests/Issues/Issue2255_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2255_Tests.cs
@@ -90,6 +90,39 @@ public class Issue2255_Tests
         }
     }
 
+    [TypeConverter(typeof(CollidingKeyConverter))]
+    public sealed class CollidingKey
+    {
+        public CollidingKey(int number)
+        {
+            Number = number;
+        }
+
+        public int Number { get; }
+    }
+
+    public sealed class CollidingKeyConverter : TypeConverter
+    {
+        public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
+        {
+            return destinationType == typeof(string) || base.CanConvertTo(context, destinationType);
+        }
+
+        public override object ConvertTo(
+            ITypeDescriptorContext context,
+            CultureInfo culture,
+            object value,
+            Type destinationType)
+        {
+            if (destinationType == typeof(string) && value is CollidingKey key)
+            {
+                return key.Number == 1 ? "same" : "SAME";
+            }
+
+            return base.ConvertTo(context, culture, value, destinationType);
+        }
+    }
+
     private class Row
     {
         public int Id { get; set; }
@@ -172,6 +205,21 @@ public class Issue2255_Tests
         Assert.Equal(2, document.Count);
         Assert.Equal("first value", document["first:42"].AsString);
         Assert.Equal("second value", document["second:42"].AsString);
+    }
+
+    [Fact]
+    public void Dictionary_key_conversion_should_reject_case_insensitive_collisions()
+    {
+        var values = new Dictionary<CollidingKey, string>
+        {
+            [new CollidingKey(1)] = "first",
+            [new CollidingKey(2)] = "second"
+        };
+
+        var exception = Record.Exception(() => new BsonMapper().Serialize(values));
+
+        Assert.NotNull(exception);
+        Assert.Contains("same", exception.Message, StringComparison.OrdinalIgnoreCase);
     }
 
     [Fact]
--- a/LiteDB.Tests/Issues/Issue2255_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2255_Tests.cs
@@ -1,0 +1,84 @@
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using Xunit;
+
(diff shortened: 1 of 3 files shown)


JKamsker:

Reviewed exact head `e8d9b3755d0ba041a8758e846a7278f0a733324d`. Three type-contract regressions survived reproduction.

### 1. P2 — ordinary sbyte[] changes from Binary to Array

~~~csharp
class Record
{
    public int Id { get; set; }
    public sbyte[] Values { get; set; }
}
~~~

Because of CLR array equivalence, an ordinary `sbyte[]` previously followed the byte-array Binary path. The new exact-type check sends it through enumerable serialization instead:

~~~text
base: BSON Binary {"$binary":"gP8Bfw=="}
head: BSON Array  [-128,-1,1,127]
~~~

Mixed database reproduction:

~~~text
base-written row: Binary
head-written row: Array
typed reads: both visible
indexed Values.Contains(-1): only 1 of 2 rows visible
~~~

Code: [BsonMapper.Serialize.cs lines 80–83](https://github.com/litedb-org/LiteDB/blob/e8d9b3755d0ba041a8758e846a7278f0a733324d/LiteDB/Client/Mapper/BsonMapper.Serialize.cs#L80-L83).

**Suggested direction:** bypass Binary only when the actual array element type is an enum, rather than requiring runtime type exactly byte[].

### 2. P2 — non-generic collection members become write-only

~~~csharp
enum ByteE : byte { A = 1, B = 2 }

class Holder
{
    public int Id { get; set; }
    public System.Collections.IEnumerable Values { get; set; }
}

collection.Insert(new Holder
{
    Values = new[] { ByteE.A, ByteE.B }
});                               // succeeds

collection.FindById(id);           // throws InvalidCtor
~~~

The runtime enum array is written as a BSON Array, but the declared non-generic contract provides no element type. Reading tries to instantiate `IEnumerable` or abstract `System.Array`.

The same failure affects non-generic `IList`, `ICollection`, and `System.Array`.

**Suggested direction:** retain Binary when the declared contract cannot recover an element type, or deserialize supported non-generic contracts into object[]/ArrayList.

### 3. P3 — an explicit byte[] contract can fail to round-trip

This legal CLR cast exposes a byte-backed enum array as byte[]:

~~~csharp
ByteE[] enums = { ByteE.A, ByteE.B };
byte[] bytes = (byte[])(Array)enums;

var bson = mapper.Serialize<byte[]>(bytes);
mapper.Deserialize<byte[]>(bson);
~~~

Base emits Binary and round-trips. Head emits an Array of enum names; the read then throws because the Array's raw `List<BsonValue>` cannot be cast to byte[].

**Suggested direction:** an explicitly declared byte[] contract should retain Binary behavior.

The intended concrete enum-array case works, but mixed Binary/Array rows already require a rewrite-and-reindex migration for array queries. These additional regressions should be resolved before introducing that representation split.

PR #2757, opened 2026-06-10
Fix #2211: allow equals signs in database file paths

### LINKED ISSUE #2211: [BUG] Database paths/names with "=" sign are not welcome.
Database paths/names with "=" sign are not welcome.
This is probably a feature but it bites. It's probably the design flaw.

`LiteDatabase` and `ConnectionString` constructors accept both database paths
and connection strings. This is "convenient" and usually works fine in test
environments with some usual database paths used as arguments.

Then an app goes to production where any valid paths are possible, including
paths with "=", a valid directory and file name character. These path are
problematic.

Please find attached 3 XUnit tests.

- `UsingFilePathDirectlyFails` shows how `LiteDatabase` fails with a valid file path.
- `UsingFilePathConnectionStringFails` shows how `LiteDatabase` fails with `ConnectionString` with a valid file path.
- `Workaround` shows probably the only reliable way with `ConnectionString` default constructor.

This issue makes using file paths in `LiteDatabase` and `ConnectionString` not
needed in properly designed applications.

Possible solution? For example, instead ambiguous constructors add safe static
`Create` methods:

- `LiteDatabase.CreateFromPath(string path)`
- `ConnectionString.CreateFromPath(string path)`

[_220630_kq-equal-sign-issue.zip](https://github.com/mbdavid/LiteDB/files/9020266/_220630_kq-equal-sign-issue.zip)

JKamsker:
Fixes #2211.

## Summary
- Treats raw strings containing `=` as filenames unless the first token is a known LiteDB connection-string key.
- Allows valid filesystem paths such as `test=1.LiteDB` to be passed directly to `ConnectionString` and `LiteDatabase`.
- Preserves existing `filename=...` and other key/value connection-string forms.

## TDD / verification
- Added Issue2211 tests based on the reporter attachment for direct path and `ConnectionString` construction.
- Added a guard that `filename=sample=1.db` still parses as a connection string.
- Verified targeted issue tests, existing `ConnectionString_Tests`, and the full `net8.0` test project.

## Compatibility note
No on-disk format change. This only adjusts constructor parsing for ambiguous strings.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring memory profiles, cache size, and transaction page limits through connection strings.

* **Bug Fixes**
  * Improved handling of database filenames and paths containing `=`.
  * Rejected malformed or ambiguous connection strings instead of treating them as filenames.
  * Improved support for quoted filenames, empty passwords, and trailing whitespace.

* **Documentation**
  * Added guidance on connection-string classification, quoting, supported options, and avoiding ambiguous input.

### DIFF
--- a/LiteDB.Tests/Issues/Issue2211_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2211_Tests.cs
@@ -1,0 +1,85 @@
+using System;
+using System.IO;
+using Xunit;
+
+namespace LiteDB.Tests.Issues
+{
+    public class Issue2211_Tests
+    {
+        [Fact]
+        public void ConnectionString_Should_Treat_Path_With_Equal_Sign_As_Filename()
+        {
+            var filename = GetTempDatabasePathWithEqualSign();
+
+            try
+            {
+                var connectionString = new ConnectionString(filename);
+
+                Assert.Equal(filename, connectionString.Filename);
+            }
+            finally
+            {
+                DeleteDatabaseFiles(filename);
+            }
+        }
+
+        [Fact]
+        public void LiteDatabase_Should_Open_File_Path_With_Equal_Sign()
+        {
+            var filename = GetTempDatabasePathWithEqualSign();
+
+            try
+            {
+                using (var db = new LiteDatabase(filename))
+                {
+                    var collection = db.GetCollection<Data>("data");
+
+                    Assert.Equal(0, collection.Count());
+                }
+            }
+            finally
+            {
+                DeleteDatabaseFiles(filename);
+            }
+        }
+
+        [Fact]
+        public void ConnectionString_Should_Still_Parse_Filename_Key()
+        {
+            var connectionString = new ConnectionString("filename=sample=1.db");
+
+            Assert.Equal("sample=1.db", connectionString.Filename);
+        }
+
+        private static string GetTempDatabasePathWithEqualSign()
+        {
+            var filename = "litedb-" + Guid.NewGuid().ToString("d").Substring(0, 5) + "=issue2211.db";
+
+            return Path.Combine(Path.GetTempPath(), filename);
+        }
+
+        private static void DeleteDatabaseFiles(string filename)
+        {
+            DeleteIfExists(filename);
+            DeleteIfExists(Path.Combine(
+                Path.GetDirectoryName(filename),
+                Path.GetFileNameWithoutExtension(filename) + "-log" + Path.GetExtension(filename)));
+            DeleteIfExists(Path.Combine(
+                Path.GetDirectoryName(filename),
+                Path.GetFileNameWithoutExtension(filename) + "-tmp" + Path.GetExtension(filename)));
+        }
+
+        private static void DeleteIfExists(string filename)
+        {
+            if (File.Exists(filename))
+            {
+                File.Delete(filename);
+            }
+        }
+
+        public class Data
+        {
+            public int Id { get; set; }
+        }
+    }
+}
--- a/LiteDB.Tests/Issues/Issue2211_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2211_Tests.cs
@@ -51,6 +51,16 @@ namespace LiteDB.Tests.Issues
             Assert.Equal("sample=1.db", connectionString.Filename);
         }
 
+        [Fact]
+        public void ConnectionString_Should_Parse_Custom_Key_Before_Filename_Key()
+        {
+            var connectionString = new ConnectionString("tenant=acme;filename=sample=1.db;readonly=true");
+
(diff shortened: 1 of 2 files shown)


JKamsker:

Reviewed exact head `d03f83e1578a2543111dd09b873f61cab5f66631`. Two parser regressions survived reproduction.

### 1. P1 — a malformed connection string silently becomes a filename

Suppose `encrypted.db` already exists and is encrypted with password `abc=def`:

~~~csharp
using var db = new LiteDatabase(
    "filename=encrypted.db;password=abc=def;readonly=");
~~~

The empty final value makes key/value parsing throw. The new blanket catch then says “this must be a raw filename” and produces:

~~~text
Filename = "filename=encrypted.db;password=abc=def;readonly="
Password = null
ReadOnly = false
~~~

Observed behavior:

~~~text
real encrypted.db:
  unchanged and never opened

new file:
  filename=encrypted.db;password=abc=def;readonly=
  plaintext
  writable
  password exposed in the filename
~~~

Before this PR the malformed configuration threw before any file was created. The same fallback applies to trailing `;bogus`, doubled separators, and empty password/upgrade/readonly options.

Code: [ConnectionString.cs lines 73–75 and 107–114](https://github.com/litedb-org/LiteDB/blob/d03f83e1578a2543111dd09b873f61cab5f66631/LiteDB/Client/Structures/ConnectionString.cs#L107-L114).

**Suggested direction:** parsing needs three results: raw path, valid connection string, and invalid connection string. Once recognized option syntax is present, propagate parse errors instead of falling back to a path.

### 2. P2 — custom-only key/value strings also become filenames

~~~csharp
var cs = new ConnectionString(
    "tenant=acme;region=west");
~~~

~~~text
base:
  Filename = ""
  cs["tenant"] = "acme"

head:
  Filename = "tenant=acme;region=west"
  cs["tenant"] = null
~~~

At the database boundary:

~~~csharp
new LiteDatabase("tenant=acme;region=west");
~~~

Base rejects the missing data source. Head creates an 8 KiB database literally named `tenant=acme;region=west`.

A typo such as `filenam=production.db` therefore succeeds by opening a new incorrectly named database instead of reporting a configuration error. Custom secret values can also leak into filenames.

Code: [ConnectionString.cs lines 116–119](https://github.com/litedb-org/LiteDB/blob/d03f83e1578a2543111dd09b873f61cab5f66631/LiteDB/Client/Structures/ConnectionString.cs#L116-L119).

**Suggested direction:** absence of a built-in key cannot prove that `foo=bar` is a path. Preserve the established custom-option behavior and add an explicit raw-path API or another unambiguous syntax.

Normal explicit `filename=...` paths, encrypted files, and v4 upgrades passed. The blocker is specifically silent wrong-database selection and loss of encryption/read-only intent on invalid or custom syntax.

JKamsker (commented):
Re-checked the fix commits: CI is green, and the equals-sign paths from #2211 now open correctly. One blocker is left before merging:

**A path followed by options is taken as a plain filename.** `C:\data\my.db;password=secret` and `data/my.db;readonly=true` are both treated as file names on this head, so the password and read-only settings are ignored and the password ends up in the file name. On dev these strings are rejected. The cause is that the "contains a folder separator, so it's a path" shortcut runs before the check for `;`.

**Fix:** skip that shortcut when the string contains `;`, and add both inputs to the malformed-connection-string test.

Not blocking, but it needs a line in the release notes: a single `key=value` with an unknown key and no `;` (e.g. the typo `filenam=production.db`, or `tenant=acme`) is now opened as a file path instead of failing. That follows from how the fix tells paths and connection strings apart.

JKamsker:
Follow-up after #2772 was merged: `LooksLikeKeyValueConnectionString` checks the first key against a fixed list (`filename`, `connection`, `password`, `initial size`, `readonly`, `upgrade`, `auto-rebuild`, `collation`). Please add #2772's new keys, `memory profile`, `cache size` and `transaction pages`. Strings that contain `;` aren't affected, since they're always parsed as connection strings. But a lone `cache size=64MB` without a filename would currently be opened as a file with that name, instead of being rejected like the other built-in keys.

JKamsker (commented):
Merge-readiness update for `45c10ce10b67`.

Changes needed before merge, consolidating the outstanding parser feedback:

- Recognize `memory profile`, `cache size` and `transaction pages` in the built-in option list. A settings-only string such as `cache size=64MB` must preserve the parsed option rather than becoming a database filename.
- Preserve rejection of malformed path-plus-options strings. The path-separator shortcut currently runs before the semicolon check and can treat the entire input as a filename, silently ignoring its options.
- Add tests for all three memory settings, malformed path-plus-options inputs, and valid filenames containing equals signs. Document the intended treatment of unknown single-key input and coordinate that contract with #944 and #2745.

These are the remaining blockers; the earlier fixes to equals-sign filenames are acknowledged.

Issue #2405, opened 2024-01-08
Error when trying to update from v4 to v5: Cannot insert duplicate key in unique index


LionAM1:
Hello,
I'm trying to update LiteDB from Version 4.1.4 to 5.0.17. However, when trying to read some files, I get an error that they contain items with a duplicate key. I checked multiple files with LiteDbExplorer and they actually contain entries with duplicate keys - although I have no idea how this could happen (I always used Collection.Upsert to add new entries).

Using version 4.1.4, I did not get an exception when opening the file and everything seemed to work - so I didn't notice the issue. However, when trying to open with 5.0.17, I immediately get the exception and cannot open the file at all. Is there a way to ignore the duplicate entries and try to repair the files?

I've added a sample file:
[Data.zip](https://github.com/mbdavid/LiteDB/files/13863923/Data.zip)

Kind regards
Alex

JKamsker:
Consolidated into #2823, which combines this report with the related ones and tracks the fix. Closing this as a duplicate — please follow #2823, and comment there if your case is different.

PR #2758, opened 2026-06-10
Fix #2190: avoid duplicate id element enumeration

### LINKED ISSUE #2190: [BUG]Class member without nullable causes column name missing last 3 letters
.net core 6.0 litedb 5.0.11

Class member without nullable causes column name missing last 3 letters when inserting to a strong type collection.

**Code to Reproduce**
        public class Statistic
        {
            public Int32 _Id { get; set; }
            public Int32 Played { get; set; } = 0;//not working, column name becomes "Pla" instead of "Played"
            //public Int32? Played { get; set; } = 0;//working with no issue
        }

                var col = db.GetCollection<Dos.Statistic>("dos_statistic");
                col.Insert(new Dos.Statistic { _Id = id });

results in database:
{
  "_id": 1,
  "Pla": 0,
}

I also tried other combination:
        public class Statistic
        {
            public Int32 _Id { get; set; }
            public Int32 Played { get; set; } = 0;
            public Int32 Played1 { get; set; } = 0;
        }

results in database:
{
  "_id": 1,
  "Played": 0,
  "Play": 5     //don't even know 5 comes from where
}

It's a show stopper currently in my project. Please advise.

JKamsker:
Fixes #2190.

## Summary
- Makes `BsonDocument.GetElements()` exclude `_id` using the same case-insensitive comparison as the backing dictionary.
- Prevents case-variant `_Id` properties from being emitted twice during serialization.
- Fixes debug-time buffer overflows and release-time field-name truncation such as `Played` becoming `Pla`.

## TDD / verification
- Added Issue2190 coverage for the raw `BsonDocument.GetElements()` duplicate element bug.
- Added end-to-end typed insert coverage for an entity with `_Id` and `Played` properties.
- Verified targeted issue tests, document tests, buffer-writer tests, and the full `net8.0` test project.

## Compatibility note
No valid on-disk format is changed. The affected documents either failed to serialize in debug builds or could be written corrupted in release builds; this prevents new corruption.

### DIFF
--- a/LiteDB.Tests/Issues/Issue2190_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue2190_Tests.cs
@@ -1,0 +1,49 @@
+using System.IO;
+using System.Linq;
+using Xunit;
+
+namespace LiteDB.Tests.Issues
+{
+    public class Issue2190_Tests
+    {
+        [Fact]
+        public void BsonDocument_GetElements_Should_Not_Duplicate_CaseVariant_Id_Key()
+        {
+            var doc = new BsonDocument();
+            doc["_Id"] = 1;
+            doc["Played"] = 0;
+
+            Assert.Equal(2, doc.Count);
+
+            var elements = doc.GetElements().ToArray();
+
+            Assert.Equal(2, elements.Length);
+            Assert.Equal("_id", elements[0].Key);
+            Assert.Equal("Played", elements[1].Key);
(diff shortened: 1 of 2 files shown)


gha-zund:
Is there a planned release date for this? I guess this could solve a problem I experience.

Is this releated? https://github.com/litedb-org/LiteDB/issues/2405

PR #2759, opened 2026-06-10
Fix #1929: deserialize System.Index values

### LINKED ISSUE #1929: [BUG] Deserialization of struct 'Index' causes an NullReferenceException
**Version**
5.0.10

**Describe the bug**
Loading a document causes an NullReferenceException during deserialization if the document contains a property of type 'Index'.

**Code to Reproduce**
```csharp
using LiteDB;
using System;
using Xunit;

namespace LiteDb.Tests {
    public class LiteDbTest {
        [Fact]
        public void FindById_DeserializationWithIndexType_IsSuccessfull() {
            // Arrange
            using var db = new LiteDatabase(":memory:");

            var person = new Person {
                Name = "Alex"
            };

            var collection = db.GetCollection<Person>();
            collection.Insert(person);

            // Act

            // Next line causes:
            // System.NullReferenceException : Object reference not set to an instance of an object.
            //Stack Trace: 
            //  lambda_method(Closure, BsonDocument)
            //  BsonMapper.Deserialize(Type type, BsonValue value)
            //  BsonMapper.DeserializeObject(EntityMapper entity, Object obj, BsonDocument value)
            //  BsonMapper.Deserialize(Type type, BsonValue value)
            //  LiteQueryable`1.< ToEnumerable > b__27_2(BsonDocument x)
            //  SelectEnumerableIterator`2.MoveNext()
            //  Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean & found)
            //  Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
            //  LiteCollection`1.FindById(BsonValue id)
            //  UnitTest1.Deserialization_Index_IsSuccessfull() line 21
            var result = collection.FindById(person.Id);

            // Assert
            Assert.Equal(result.Name, person.Name);
        }
    }

    public class Person {
        public string Id { get; set; } = Guid.NewGuid().ToString();
        public Index Selection { get; set; }  // <- Causes the exception
    }
}
```

**Expected behavior**
The struct 'Index' can be deserialized.
See: https://docs.microsoft.com/en-us/dotnet/api/system.index?view=net-5.0

**Screenshots/Stacktrace**
```
Result StackTrace:	
at lambda_method(Closure , BsonDocument )
   at LiteDB.BsonMapper.Deserialize(Type type, BsonValue value)
   at LiteDB.BsonMapper.DeserializeObject(EntityMapper entity, Object obj, BsonDocument value)
   at LiteDB.BsonMapper.Deserialize(Type type, BsonValue value)
   at LiteDB.LiteQueryable`1.<ToEnumerable>b__27_2(BsonDocument x)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
   at LiteDB.LiteCollection`1.FindById(BsonValue id)
   at LiteDb.Tests.LiteDbTest.FindById_DeserializationWithIndexType_IsSuccessfull() in C:\dev\LiteDb.Tests\LiteDb.Tests\LiteDbTest.cs:line 34
Result Message:	System.NullReferenceException : Object reference not set to an instance of an object.
```

**Additional context**
Test environment
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>

    <IsPackable>false</IsPackable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="LiteDB" Version="5.0.10" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.3" />
    <PackageReference Include="xunit" Version="2.4.1" />
    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
    <PackageReference Include="coverlet.collector" Version="3.0.2">
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
  </ItemGroup>

</Project>

```

JKamsker:
Fixes #1929.

## Summary
- Reconstructs `System.Index` from the existing BSON document shape produced by the mapper: `{ Value, IsFromEnd }`.
- Fixes deserialization returning the default `Index` value for persisted `System.Index` properties.
- Uses reflection by type name so `netstandard2.0` builds do not take a compile-time dependency on `System.Index`.

## TDD / verification
- Added net8-only Issue1929 regression coverage for a typed collection containing a `System.Index` property.
- Verified targeted issue tests, mapper tests, `netstandard2.0` library build, and the full `net8.0` test project.

## Compatibility note
No on-disk format change. The fix reads the shape already written by the mapper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed deserialization of persisted `System.Index` values.
  * Improved compatibility with different field casing and document representations.
  * Correctly restores index values, including from-end indexes.
  * Prevented unrelated types with the same name from being incorrectly treated as `System.Index`.

### DIFF
--- a/LiteDB.Tests/Issues/Issue1929_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue1929_Tests.cs
@@ -74,6 +74,25 @@ namespace LiteDB.Tests.Issues
         }
 
         [Fact]
+        public void Object_Type_Discriminator_Should_Still_Honor_The_Custom_Index_Constructor()
+        {
+            var mapper = new BsonMapper();
+            mapper.Entity<System.Index>().Ctor(document => new System.Index(
+                document["offset"].AsInt32,
+                document["fromEnd"].AsBoolean));
+            var stored = mapper.Serialize(typeof(object), System.Index.FromEnd(19)).AsDocument;
+
+            stored.Remove("Value");
+            stored.Remove("IsFromEnd");
+            stored["offset"] = 19;
+            stored["fromEnd"] = true;
+
+            var result = mapper.Deserialize(typeof(object), stored);
+
+            Assert.Equal(System.Index.FromEnd(19), Assert.IsType<System.Index>(result));
+        }
+
+        [Fact]
         public void A_User_Type_Merely_Named_System_Index_Should_Map_Normally()
         {
             var mapper = new BsonMapper();
--- a/LiteDB.Tests/Issues/Issue1929_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue1929_Tests.cs
@@ -1,6 +1,8 @@
 #if !NETFRAMEWORK
 using System;
 using System.IO;
+using System.Reflection;
+using System.Reflection.Emit;
 using Xunit;
 
 namespace LiteDB.Tests.Issues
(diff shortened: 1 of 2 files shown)


JKamsker:

Reviewed exact head `64e29e825e2829dfda39f0133503073d4e2d5344`. Two deserialization regressions survived reproduction.

### 1. P1 — the fast path bypasses supported custom System.Index mappings

Existing applications can work around immutable `Index` mapping with:

~~~csharp
mapper.Entity<Index>()
    .Field(x => x.Value, "offset")
    .Ctor(doc => new Index(
        doc["offset"].AsInt32,
        doc["IsFromEnd"].AsBoolean));
~~~

A base-created database then stores:

~~~json
{"offset":17,"IsFromEnd":true}
~~~

Results after upgrade:

~~~text
base typed read -> ^17
head raw read   -> {"offset":17,"IsFromEnd":true}
head typed read -> ^0
~~~

The new special case runs before `EntityMapper`, looks only for `Value`, receives BSON Null, and converts that to zero. If both fields are renamed, the same read throws NullReferenceException instead.

This affects nullable values, nested objects, collections, dictionaries, and discriminated object values. It also bypasses configured constructors/type instantiators.

Code: [BsonMapper.Deserialize.cs lines 138–140](https://github.com/litedb-org/LiteDB/blob/64e29e825e2829dfda39f0133503073d4e2d5344/LiteDB/Client/Mapper/BsonMapper.Deserialize.cs#L138-L140) and [300–317](https://github.com/litedb-org/LiteDB/blob/64e29e825e2829dfda39f0133503073d4e2d5344/LiteDB/Client/Mapper/BsonMapper.Deserialize.cs#L300-L317).

**Suggested direction:** honor explicitly configured entity constructors and resolved MemberMapper field names before using the built-in fallback. Require both fields instead of interpreting a missing magnitude as zero.

### 2. P2 — matching only FullName hijacks unrelated types

The new check is effectively:

~~~csharp
type.FullName == "System.Index"
~~~

Full names do not include the assembly. A user type or polyfill named `System.Index` in another assembly is therefore treated as the BCL struct.

Independent repro defined:

~~~text
System.Index, UserDomain
  public parameterless constructor
  writable Value and IsFromEnd properties
~~~

~~~text
base deserialization -> succeeds
head deserialization -> MissingMethodException
~~~

Existing databases storing that type become unreadable after upgrade. The discriminator path has the same problem.

**Suggested direction:** compare actual Type identity with the known runtime BCL `System.Index`, ideally a direct `typeof(Index)` comparison, rather than comparing names.

Default `System.Index` storage remains compatible. The merge blocker is the newly bypassed public mapping configuration for databases already using the supported workaround.

JKamsker (commented):
Thanks! The Aug 11 fixes address the main review points: only CoreLib's `System.Index` is matched, and a configured instantiator or constructor runs first. Two things before merging:

1. A missing field still silently becomes 0: `GetSystemIndexField` returns `BsonValue.Null`, and `AsInt32` turns that into 0. Please require both fields (or fall back to the entity path), and resolve the names through `MemberMapper` so mappings that only rename fields still work.
2. `System.Range` is affected the same way; it reads back as `0..0`. Please cover `Start`/`End` too, or we'll open a follow-up so #1929 isn't closed half-fixed.

PR #101, opened 2016-01-14
Use UtcNow in the FileDiskService.TryExec

remember664:
Use `DateTime.UtcNow` for timeout checking, because `DateTime.UtcNow` is much more faster than `DateTime.Now`

#### Performance comparison (on my system)
`DateTime.Now`: 1071 ms
```C#
for (int i = 0; i < 1000000; i++)
{
  var timer = DateTime.Now.Add(TimeSpan.FromMinutes(1));
  var tmp = DateTime.Now < timer;
}
```
`DateTime.Now`: 37 ms
```C#
for (int i = 0; i < 1000000; i++)
{
  var timer = DateTime.UtcNow.Add(TimeSpan.FromMinutes(1));
  var tmp = DateTime.UtcNow < timer;
}
```
`DateTime.Now`: 526ms
```C#
var timer = DateTime.Now.Add(TimeSpan.FromMinutes(1));
for (int i = 0; i < 1000000; i++)
{
  var tmp = DateTime.Now < timer;
}
```

`DateTime.UtcNow`: 10ms
```C#
var timer = DateTime.UtcNow.Add(TimeSpan.FromMinutes(1));
for (int i = 0; i < 1000000; i++)
{
  var tmp = DateTime.UtcNow < timer;
}
```
Related [blog post](http://blogs.msdn.com/b/kirillosenkov/archive/2012/01/10/datetime-utcnow-is-generally-preferable-to-datetime-now.aspx)

PR #2762, opened 2026-07-28
Fix #1759: alignment-safe ToBytes for Int64/UInt64 (SIGBUS/BUS_ADRALN on 32-bit ARM)

### LINKED ISSUE #1759: [BUG] BufferExtensions.ToBytes() methods crash on Android ARMv7 with Unity IL2CPP

papajames89:
Fixes #1759.

## Problem
`BufferExtensions.ToBytes(Int64/UInt64)` wrote 8 bytes through a single pointer store:

```csharp
public static unsafe void ToBytes(this Int64 value, byte[] array, int startIndex)
{
    fixed (byte* ptr = &array[startIndex]) { *(Int64*)ptr = value; }
}
```

On 32-bit ARM (`armeabi-v7a`) that store is lowered to `STRD`/`STM`, which require a
word-aligned address. When `startIndex` is not word-aligned the CPU raises an alignment
fault — **`SIGBUS`, code `BUS_ADRALN`** — and the process dies. Reported with Unity
IL2CPP; also reproduced with .NET for Android using Mono LLVM AOT
(`RunAOTCompilation` + `EnableLLVM`).

It is not an edge case. `EnginePragmas.UpdateBuffer` writes the `LIMIT_SIZE` pragma
(default `long.MaxValue`) at the fixed page offset `P_LIMIT_SIZE = 101`; because
`101 % 4 == 1` the destination is always unaligned, so every header-page flush crashes.

## Fix
The 8-byte stores are done byte by byte, which has no alignment requirement:

- **Byte order follows the platform** (`BitConverter.IsLittleEndian`), so the bytes are
  identical to what the pointer store produced. The on-disk layout is unchanged, which
  matters because the matching readers (`ReadInt64`/`ReadUInt64`/`ReadDouble`) use
  native-endian `BitConverter`.
- The stores run inside `unchecked`, so the narrowing casts behave like the removed
  pointer store even when the library is compiled with `CheckForOverflowUnderflow=true`.
- `Int64` delegates to the `UInt64` overload; `Double` already routed through `UInt64`.
- The 16/32-bit overloads are left alone: single-word `STR`/`STRH` tolerate unaligned
  access on ARMv7, only the 8-byte `STRD`/`STM` path faults.

No new dependencies.

## Tests
`LiteDB.Tests/Issues/Issue1759_Tests.cs` pins down the two properties that must hold on
every platform, at unaligned offsets (0, 1, 2, 3, 5, 7):

- the written bytes equal `BitConverter.GetBytes(value)` (native order), and
- a write/read round trip through `BufferSlice` returns the original value
  (`Int64`, `Double`, `DateTime`).

The round-trip test is architecture-independent: it fails on any platform where writer
and reader disagree about byte order, not just on little-endian hosts. The alignment
fault itself cannot be observed on x86/x64, since unaligned access is allowed there.

## Verification
- Full suite on `net8.0`: 299 passed / 6 skipped, no regressions (baseline `dev`: 275/6,
  plus the 24 new cases here).
- Same suite with `-p:CheckForOverflowUnderflow=true`: all green; the IL for
  `ToBytes(UInt64)` contains no `conv.ovf` while the rest of the assembly does, so the
  `unchecked` block is doing its job in a genuinely checked build.
- On-disk compatibility: a database written by `dev` is read back by this branch with all
  values intact (`long`, `ulong`, `double`, `DateTime`), and vice versa.
- Machine code, `android-arm` + AOT + LLVM: in a build of unpatched `dev` the pattern
  "array element address → 8-byte multi-word store" appears 8 times in
  `libaot-LiteDB.dll.so`, including `stmne r1, {r0, r3}` directly after
  `add r1, r1, #101` (the `LIMIT_SIZE` write). With this patch that pattern is gone
  (0 occurrences) — the byte stores are not merged back into `STRD`/`STM`.
- On a physical 32-bit ARM device, the earlier byte-wise version of this patch turned a
  reproducible `SIGBUS/BUS_ADRALN` on the first header flush into a clean run with LLVM
  AOT still enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved 64-bit integer and floating-point serialization at unaligned buffer offsets.
  * Prevented potential failures on platforms sensitive to unaligned memory access.
  * Preserved native byte-order behavior across supported platforms.

* **Tests**
  * Added coverage across multiple offsets for `Int64`, `UInt64`, and `Double`.
  * Added round-trip validation for 64-bit integers, floating-point values, and dates.

### DIFF
--- a/LiteDB.Tests/Issues/Issue1759_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue1759_Tests.cs
@@ -1,65 +1,90 @@
 using System;
 using FluentAssertions;
-using LiteDB;
 using Xunit;
 
 namespace LiteDB.Tests.Issues
 {
     /// <summary>
-    /// #1759 - BufferExtensions.ToBytes(Int64/UInt64/Double) used an unaligned
-    /// *(long*)ptr = value store. On 32-bit ARM (armeabi-v7a) this crashes with
-    /// SIGBUS / BUS_ADRALN under IL2CPP or Mono-LLVM Release, because the compiler
-    /// emits STRD/STM which require a word-aligned address. The write must be
-    /// alignment-safe (byte-wise).
+    /// #1759 - ToBytes(Int64/UInt64/Double) stored 8 bytes through a single
+    /// `*(long*)ptr = value` write. On 32-bit ARM (armeabi-v7a) that is lowered to
+    /// STRD/STM, which fault with SIGBUS/BUS_ADRALN when the destination is not
+    /// word-aligned, so writing e.g. the LIMIT_SIZE pragma (page offset 101) killed
+    /// the process. The store is now done byte by byte, which has no alignment
+    /// requirement.
     ///
-    /// The alignment fault itself cannot be observed on x86/x64 CI (those tolerate
-    /// unaligned access), so these tests assert byte-level correctness at unaligned
-    /// offsets to guard against a regression to the pointer-cast implementation.
+    /// The alignment fault itself cannot be observed on x86/x64 (unaligned access is
+    /// allowed there), so these tests pin down the two properties that must hold on
+    /// every platform: the bytes stay in the platform's native order, and the value
+    /// survives a write/read round trip through the buffer at unaligned offsets.
     /// </summary>
     public class Issue1759_Tests
     {
+        public static TheoryData<int> UnalignedOffsets => new TheoryData<int> { 0, 1, 2, 3, 5, 7 };
+
         [Theory]
-        [InlineData(0)]
-        [InlineData(1)]
-        [InlineData(2)]
-        [InlineData(3)]
-        public void ToBytes_Int64_is_correct_at_unaligned_offset(int offset)
+        [MemberData(nameof(UnalignedOffsets))]
+        public void ToBytes_Int64_keeps_native_byte_order(int offset)
         {
-            var value = long.MaxValue; // 0x7FFFFFFFFFFFFFFF - the value seen in real crashes
-            var buffer = new byte[offset + 8];
+            foreach (var value in new[] { long.MaxValue, long.MinValue, -1L, 1L, 0x0102030405060708L })
+            {
+                var buffer = new byte[offset + 8];
 
-            value.ToBytes(buffer, offset);
+                value.ToBytes(buffer, offset);
 
-            BitConverter.ToInt64(buffer, offset).Should().Be(value);
+                // BitConverter is native-endian, and so are the matching readers, so the
+                // bytes written must be byte-for-byte what BitConverter produces.
+                buffer.AsSpan(offset, 8).ToArray()
+                    .Should().Equal(BitConverter.GetBytes(value), "value {0} at offset {1}", value, offset);
+            }
         }
 
         [Theory]
-        [InlineData(0)]
-        [InlineData(1)]
-        [InlineData(2)]
-        [InlineData(3)]
-        public void ToBytes_UInt64_is_correct_at_unaligned_offset(int offset)
+        [MemberData(nameof(UnalignedOffsets))]
+        public void ToBytes_UInt64_keeps_native_byte_order(int offset)
         {
-            var value = ulong.MaxValue - 12345UL;
-            var buffer = new byte[offset + 8];
+            foreach (var value in new[] { ulong.MaxValue, ulong.MinValue, 1UL, 0x0102030405060708UL })
+            {
+                var buffer = new byte[offset + 8];
 
-            value.ToBytes(buffer, offset);
+                value.ToBytes(buffer, offset);
 
-            BitConverter.ToUInt64(buffer, offset).Should().Be(value);
+                buffer.AsSpan(offset, 8).ToArray()
+                    .Should().Equal(BitConverter.GetBytes(value), "value {0} at offset {1}", value, offset);
+            }
         }
 
         [Theory]
-        [InlineData(0)]
-        [InlineData(1)]
-        [InlineData(3)]
-        public void ToBytes_Double_is_correct_at_unaligned_offset(int offset)
+        [MemberData(nameof(UnalignedOffsets))]
+        public void ToBytes_Double_keeps_native_byte_order(int offset)
         {
-            var value = Math.PI;
-            var buffer = new byte[offset + 8];
+            foreach (var value in new[] { Math.PI, 0d, -1.5d, double.MaxValue, double.MinValue })
+            {
+                var buffer = new byte[offset + 8];
 
-            value.ToBytes(buffer, offset);
+                value.ToBytes(buffer, offset);
 
-            BitConverter.ToDouble(buffer, offset).Should().Be(value);
+                buffer.AsSpan(offset, 8).ToArray()
+                    .Should().Equal(BitConverter.GetBytes(value), "value {0} at offset {1}", value, offset);
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(UnalignedOffsets))]
+        public void BufferSlice_write_read_round_trips_at_unaligned_offset(int offset)
+        {
+            // Exercises the real writer + reader pair, so it fails on any platform where
+            // the two disagree about byte order - not just on little-endian hosts.
+            var slice = new BufferSlice(new byte[offset + 8 + 8], offset, 16);
+
+            slice.Write(long.MaxValue, 1);
+            slice.ReadInt64(1).Should().Be(long.MaxValue);
+
+            slice.Write(Math.PI, 1);
+            slice.ReadDouble(1).Should().Be(Math.PI);
+
+            var utcNow = DateTime.UtcNow;
+            slice.Write(utcNow, 1);
+            slice.ReadDateTime(1).Should().BeCloseTo(utcNow, TimeSpan.FromMilliseconds(1));
         }
     }
 }
--- a/LiteDB.Tests/Issues/Issue1759_Tests.cs
+++ b/LiteDB.Tests/Issues/Issue1759_Tests.cs
@@ -1,0 +1,65 @@
+using System;
+using FluentAssertions;
+using LiteDB;
+using Xunit;
+
+namespace LiteDB.Tests.Issues
+{
+    /// <summary>
+    /// #1759 - BufferExtensions.ToBytes(Int64/UInt64/Double) used an unaligned
+    /// *(long*)ptr = value store. On 32-bit ARM (armeabi-v7a) this crashes with
+    /// SIGBUS / BUS_ADRALN under IL2CPP or Mono-LLVM Release, because the compiler
+    /// emits STRD/STM which require a word-aligned address. The write must be
+    /// alignment-safe (byte-wise).
+    ///
+    /// The alignment fault itself cannot be observed on x86/x64 CI (those tolerate
+    /// unaligned access), so these tests assert byte-level correctness at unaligned
+    /// offsets to guard against a regression to the pointer-cast implementation.
+    /// </summary>
+    public class Issue1759_Tests
+    {
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(2)]
+        [InlineData(3)]
+        public void ToBytes_Int64_is_correct_at_unaligned_offset(int offset)
+        {
+            var value = long.MaxValue; // 0x7FFFFFFFFFFFFFFF - the value seen in real crashes
+            var buffer = new byte[offset + 8];
+
+            value.ToBytes(buffer, offset);
+
+            BitConverter.ToInt64(buffer, offset).Should().Be(value);
+        }
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(2)]
+        [InlineData(3)]
+        public void ToBytes_UInt64_is_correct_at_unaligned_offset(int offset)
+        {
+            var value = ulong.MaxValue - 12345UL;
+            var buffer = new byte[offset + 8];
+
+            value.ToBytes(buffer, offset);
+
+            BitConverter.ToUInt64(buffer, offset).Should().Be(value);
+        }
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(3)]
+        public void ToBytes_Double_is_correct_at_unaligned_offset(int offset)
+        {
+            var value = Math.PI;
+            var buffer = new byte[offset + 8];
+
+            value.ToBytes(buffer, offset);
+
+            BitConverter.ToDouble(buffer, offset).Should().Be(value);
+        }
+    }
+}
--- a/LiteDB/Utils/Extensions/BufferExtensions.cs
+++ b/LiteDB/Utils/Extensions/BufferExtensions.cs
@@ -115,12 +115,20 @@ namespace LiteDB
         /// <summary>
         /// Copy Int64 bytes direct into buffer
         /// </summary>
-        public static unsafe void ToBytes(this Int64 value, byte[] array, int startIndex)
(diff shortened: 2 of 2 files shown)


JKamsker:

Reviewed exact head `a5a94cf42696f90e107b99d7822b7ed8eba8a23f`. The ARM32 alignment fix is needed, but two portability/build defects survived reproduction.

### 1. P1 — hard-coded little-endian writes disagree with native-endian readers

The removed pointer stores wrote in native byte order. The new code always emits little-endian bytes, while the readers still use native-endian `BitConverter`.

On supported big-endian s390x:

~~~text
value:                0x0102030405060708

base/native bytes:    01 02 03 04 05 06 07 08
PR-head bytes:        08 07 06 05 04 03 02 01

unchanged BE reader:  0x0807060504030201
~~~

This affects persisted Int64, Double, DateTime, mapped UInt64, index keys, header CreationTime, and LIMIT_SIZE. For example, default `long.MaxValue` is read back as `-129`.

Existing-database failure shape:

~~~text
old big-endian database opens correctly
PR head rewrites one page/header
new 64-bit fields are little-endian
old fields remain big-endian
PR head and a downgraded version both misread the rewritten fields
WAL/checkpoint copy the mixed bytes; they do not repair them
~~~

Code: [BufferExtensions.cs lines 124–131](https://github.com/litedb-org/LiteDB/blob/a5a94cf42696f90e107b99d7822b7ed8eba8a23f/LiteDB/Utils/Extensions/BufferExtensions.cs#L124-L131) and [163–170](https://github.com/litedb-org/LiteDB/blob/a5a94cf42696f90e107b99d7822b7ed8eba8a23f/LiteDB/Utils/Extensions/BufferExtensions.cs#L163-L170). The unchanged readers are native-endian.

This is an actual supported target: .NET 8 supports Linux s390x, the runtime marks it big-endian, and LiteDB's AnyCPU net8.0/netstandard2.0 targets do not exclude it.

**Suggested direction:** keep the alignment-safe byte writes but preserve native ordering by branching on `BitConverter.IsLittleEndian`, or safely copy `BitConverter.GetBytes(value)`. Converting the whole database format to little-endian would require updating every reader/writer together plus a versioned migration.

### 2. P2 — narrowing casts throw when overflow checking is enabled

The new byte extraction uses casts such as:

~~~csharp
array[startIndex++] = (byte)value;
~~~

Their behavior depends on the compiler's checked-overflow setting. Exact reproduction:

~~~bash
dotnet test LiteDB.Tests/LiteDB.Tests.csproj \
  -f net10.0 -c Release \
  -p:LiteDBTestsEnableNet10=true \
  -p:CheckForOverflowUnderflow=true \
  --filter FullyQualifiedName~Issue1759_Tests
~~~

Result on exact head:

~~~text
11 tests
11 failures
System.OverflowException

Int64  -> BufferExtensions.cs line 124
UInt64 -> BufferExtensions.cs line 163
Double -> UInt64 overload, line 163
~~~

An end-to-end checked build also fails during normal database creation:

~~~text
ToBytes(Int64)
-> Write(DateTime)
-> HeaderPage constructor
-> OverflowException
~~~

CreationTime ticks, long.MaxValue, negative values, and nearly every Double exceed one byte. The removed pointer stores did not perform narrowing conversions.

**Suggested direction:** wrap byte extraction in `unchecked` and add a checked-build regression job.

On little-endian default builds, byte-for-byte comparisons matched the base and the 11 normal tests passed. The two fixes above can preserve that ARM32 alignment benefit without breaking s390x or checked builds.

papajames89:
Thanks — both findings are valid, I reproduced them and fixed them in 95690963.

### P1 — endianness: confirmed
The readers are native-endian (`ReadInt64`/`ReadUInt64`/`ReadDouble` are thin
`BitConverter` wrappers) and there is no endianness handling anywhere in the library, so
the format is native-endian and my hard-coded little-endian writes broke that pairing.
Modelling big-endian semantics (pointer store and `BitConverter` both MSB-first) gives
exactly the numbers you quoted:

```
value 0x0102030405060708
  base   bytes=01 02 03 04 05 06 07 08 -> read 72623859790382856      OK
  PRhead bytes=08 07 06 05 04 03 02 01 -> read 578437695752307201     CORRUPT
value 0x7fffffffffffffff
  base   bytes=7f ff ff ff ff ff ff ff -> read 9223372036854775807    OK
  PRhead bytes=ff ff ff ff ff ff ff 7f -> read -129                   CORRUPT
```

Fixed by branching on `BitConverter.IsLittleEndian`, so the bytes written are identical to
the ones the removed pointer store produced and the on-disk layout is untouched. Verified
by cross-reading databases: one written by `dev` reads back on this branch with all values
intact (`long`, `ulong`, `double`, `DateTime`), and one written by this branch reads back
on `dev`.

### P2 — checked builds: confirmed
`unchecked` added around the stores. Verified two ways: the tests pass with
`-p:CheckForOverflowUnderflow=true`, and the IL for `ToBytes(UInt64)` contains no
`conv.ovf` while the rest of the assembly contains 143 of them, so the block is effective
in a genuinely checked build.

One detail in case it helps anyone re-running your command: with a warm `obj/bin` the
property did not reach `LiteDB.dll` for me — the IL still had `conv.u1` and the 11 tests
passed. After clearing `LiteDB/obj LiteDB/bin LiteDB.Tests/obj LiteDB.Tests/bin` it
reproduced exactly as you described (11 tests, 11 failures, `OverflowException`).

### ARM32 fix still holds
Re-checked at machine-code level for `android-arm` + AOT + LLVM. In a build of unpatched
`dev`, the pattern "array element address → 8-byte multi-word store" occurs 8 times in
`libaot-LiteDB.dll.so`, including the `LIMIT_SIZE` write:

```
add   r1, r1, #101
adds  r1, r1, #16
ldrne r3, [r0, #0x24]
ldrne r0, [r0, #0x20]
stmne r1, {r0, r3}      <-- SIGBUS/BUS_ADRALN when r1 % 4 != 0
```

With this patch that pattern is gone (0 occurrences), i.e. the byte stores are not merged
back into `STRD`/`STM`.

### Tests
Rewrote them to assert native byte order (`== BitConverter.GetBytes(value)`) at unaligned
offsets plus a write/read round trip through `BufferSlice`. The round trip is the
architecture-independent one — it fails wherever writer and reader disagree about byte
order, which is what would have caught P1. Note this deliberately goes the other way from
the CodeRabbit nitpick asking for hard-coded little-endian expectations: pinning the tests
to little-endian output would bake in the very regression you reported.

Full suite on `net8.0`: 299 passed / 6 skipped, no regressions.

Happy to add a checked-build CI job as you suggested if you want it in this PR — I left
workflow files untouched for now.

JKamsker (approved):

JKamsker:
Thanks for your contribution

PR #2766, opened 2026-08-20
Added AOT related attributes, and minor changes for AOT-readiness

zbalkan:
For AOT readiness:

- Added proper attributes
- Marked IL-compliance warnings as errors
- Added AOT smoke-test project and CI action
- Made reflection-related changes

The collection-name resolver needs a caller-visible trimming contract because its default behavior reflects over runtime type interfaces and generic arguments. The PR preserves assignment syntax for custom resolvers, but replacing the original public field with a property changes the binary member shape. Consumers should therefore rebuild against the patched LiteDB package.

## Summary by CodeRabbit

* **New Features**
  * Added Native AOT-compatible source-generated BSON mappings for supported LiteDB models.
  * Added `GetGeneratedCollection<T>` for using generated mappings with CRUD, queries, indexes, and updates.
  * Added support for generated scalar, nullable, collection, dictionary, inheritance, and `DateTimeOffset` mappings.
  * Added the separate LiteDB.SourceGenerator analyzer package.

* **Documentation**
  * Added Native AOT, source-generator, package, and versioning guidance.

* **Tests**
  * Added managed, trimmed, and Native AOT coverage, including packaged-consumer validation.


JKamsker:
Pls address coderabbit reviews

zbalkan:
Done. I shouldn't have touched that `ToArray()`.

jamesshenry:
perhaps a question that belongs elsewhere but why is net8 still being targeted when it leaves support in November?

docs/AOT.md
```diff
@@ -0,0 +1,129 @@
...
+```csharp
+using LiteDB;
+
+public sealed class Customer
+{
+    public int Id { get; set; }
+    public string Name { get; set; } = string.Empty;
+}
+
+public static class AotExample
```
jamesshenry: Also, is the plan to eventually replace this manual mapping requirement with a source generator? LiteDB already has attributes that could be used as anchor points for sourcegen. I have a working experiment in my own branch

zbalkan: I was thinking of using a source generator but decided to do something minimal. Maybe in the future?

jamesshenry: if you did go that route I don't think the `LiteAotDatabase` would be necessary at all, so consumers would have much less friction to start benefiting from NativeAOT imo

zbalkan: Good. Then I'll convert this PR to a draft until then. I guess I can handle it this weekend.

JKamsker (changes requested):
Merge-readiness update for `9ca68eb5daba`.

Changes needed before merge:

- Bring changed C# files within the repository size policy without expanding legacy exceptions merely to fit this PR. The reviewed head exceeded limits in `LiteDB.AotSmokeTests/Program.cs`, `LiteDB.AotTests/SourceGeneratedMappingTests.cs`, `LiteDB.SourceGenerator/BsonSourceGenerator.cs`, `LiteDB/Client/Database/LiteQueryable.cs`, `LiteDB/Client/Mapper/Linq/LinqExpressionVisitor.cs` and `LiteDB/Document/Expression/Parser/BsonExpressionParser.cs`. Recheck the complete diff after updating to current dev, which now contains additional partial-file extractions.
- Fix the package-consumer script invocation: workflows execute `./scripts/validate-source-generator-package-consumer.sh`, but the script is stored without its executable bit. Either commit the executable mode or invoke it through bash consistently.
- Validate actual native AOT publish and execution, trimming, and a clean external consumer of the packed source-generator/library artifacts. Attach these results in addition to managed mapper tests and a solution build.
- Integrate the current .NET 10 targets and mapper extension points from #2771/#1212, preserving the warning/annotation contracts. Run the relevant CI and package checks on that updated result.

The previous 19 focused mapper tests and solution build passed; they do not establish native AOT or packaged-consumer readiness.

zbalkan:
It's going to take time. You can skip this for the nearest release. I'll finish it soon but cannot promise a time.

JKamsker:
Addressed all outstanding review feedback in `3bd60c3a` after merging current `dev` in `a4d8b43c`.

Highlights:
- split the oversized smoke-test, mapper-test, source-generator, and LINQ visitor files without growing size exceptions;
- fixed the source-generator compatibility shim, nullable numeric/Guid auto-ID generation, inherited generated-member query mapping, sample validation, and generated hint lookup;
- made all directly invoked validation scripts executable, added an AOT-job timeout, completed the net10/mapper-extension-point integration, and documented compatibility implications;
- added focused regressions for nullable auto IDs and inherited generated mapping.

Local validation:
- `python3 scripts/check-csharp-size.py` (passes; warnings only)
- `dotnet build LiteDB.sln -c Release -p:TestingEnabled=true` (passes)
- `dotnet test LiteDB.AotTests/LiteDB.AotTests.csproj -c Release --settings tests.runsettings` (77 passed)
- solution tests: .NET 8 and .NET 10 each 864 passed / 7 skipped; repro runner 18 passed; the Linux image lacks Mono for executing net462/net481, while both targets compile successfully
- `./scripts/validate-source-generator-package-consumer.sh` (clean package consumer passes trimmed and Native AOT execution)
- `./scripts/validate-aot-feature-parity.sh` (regular, trimmed single-file, and Native AOT transcripts match)
- package archive validation passes
- incremental measurement with one sample passes and verifies `LiteDbGeneratedMappings.v2.g.cs`

All eight outstanding review threads are resolved.

zbalkan:
Hi @JKamsker,

AOT is a different kind of beast to wrestle. AI tools can cause more issues than they help. Generic optimisation recommendations may fail as they will win at runtime but fail for AOT, etc. In the end, they will cause regressions.

Instead of relying on the feedback of Code Rabbit, I suggest using a second LLM, Codex, Claude or whatever, feed it with AOT-specific resources first, and make it the peer reviewer for the comments of Code Rabbit targeting AOT and trimming compatible changes. That's healthier in the long run. It can help detecting false positives better.

scripts/validate-aot-feature-parity.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+project="$repo_root/LiteDB.AotSmokeTests/LiteDB.AotSmokeTests.csproj"
+runtime_identifier="${RUNTIME_IDENTIFIER:-linux-x64}"
+output_root="${AOT_PARITY_OUTPUT_ROOT:-$repo_root/artifacts/aot-feature-parity}"
+
+rm -rf "$output_root"
```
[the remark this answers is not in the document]
JKamsker: Fixed in 4ca4bea4. Full recursive deletion remains only for the exact repository-owned default path. An override must be a real directory and a dedicated/marked workspace; cleanup removes only the regular, trimmed, and native-aot outputs/logs. A guard test confirmed an unrelated file is preserved and the script refuses the directory.

LiteDB.SourceGenerator/BsonSourceGenerator.ModelAnalysis.cs
```diff
@@ -0,0 +1,270 @@
...
+    private static ScalarConversionKind GetScalarConversionKind(ITypeSymbol type, out bool isNullableScalar)
+    {
+        isNullableScalar = false;
+        if (type is INamedTypeSymbol { OriginalDefinition.SpecialType: SpecialType.System_Nullable_T, TypeArguments.Length: 1 } nullableType)
+        {
+            isNullableScalar = true;
+            type = nullableType.TypeArguments[0];
+        }
+
+        if (type.TypeKind == TypeKind.Enum) return ScalarConversionKind.Enum;
```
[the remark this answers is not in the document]
JKamsker: Fixed in 4ca4bea4. The generator now carries each enum underlying integral kind in its equatable model and emits Int32 or Int64 serialization/deserialization accordingly, including unchecked UInt64 bit preservation. Added generated-collection regressions for long/ulong values outside Int32 and compatible Int64 BSON input; the full AOT test project passes (84/84).

LiteDB/Client/Database/LiteQueryable.cs
```diff
@@ -178,12 +197,31 @@ public ILiteQueryable<T> ThenByDescending<K>(Expression<Func<T, K>> keySelector)
...
+            var expression = this.GetExpression(keySelector);
 
             this.GroupBy(expression);
 
-            _mapper.RegisterGroupingType<K, T>();
+            if (_useGeneratedMappers)
+            {
+                IGrouping<K, T> DeserializeGrouping(BsonDocument document)
+                {
+                    var key = GeneratedScalarConverter.Convert<K>(document[LiteGroupingFieldNames.Key]);
```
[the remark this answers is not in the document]
JKamsker: Fixed in 4ca4bea4. Generated GroupBy now resolves a registered execution map for complex K, validates its options before execution, retains scalar conversion for scalar K, and rejects unmapped complex keys immediately. Added tests for mapped complex-key materialization and pre-execution anonymous-key rejection.

JKamsker:
Addressed the remaining CodeRabbit outside-diff finding in 3e1b4394: scored vector results now use the existing generated `_deserialize` delegate and only retain runtime mapper deserialization for ordinary queryables. `SourceGeneratedVectorResultTests` overrides `BsonMapper.Deserialize(Type, BsonValue)` to fail, reproduces the old path, and passes through the generated path. The same commit also replaces the generator model’s raw `ImmutableArray<PropertyDescriptor>` with an element-wise `EquatableArray<T>` wrapper; tracked second-run tests confirm the model, collected-model, and source-output stages cache while real changes invalidate them.

JKamsker:
> Hi @JKamsker,
> 
> AOT is a different kind of beast to wrestle. AI tools can cause more issues than they help. Generic optimisation recommendations may fail as they will win at runtime but fail for AOT, etc. In the end, they will cause regressions.
> 
> Instead of relying on the feedback of Code Rabbit, I suggest using a second LLM, Codex, Claude or whatever, feed it with AOT-specific resources first, and make it the peer reviewer for the comments of Code Rabbit targeting AOT and trimming compatible changes. That's healthier in the long run. It can help detecting false positives better.

@zbalkan 
Yeah, fair point - I probably overshot the scope a little while experimenting here. I'll move the broader experiments into a separate branch and keep this PR focused. Sorry for the churn 😅

And agreed regarding AI review. CodeRabbit shouldn't be authoritative for AOT-specific changes. I'll treat its comments as leads and cross-check them against AOT/trimming-specific guidance and another reviewer/model before applying them.

I still think source generators have a lot of potential for LiteDB, especially together with #2905 and the planned memory improvements - but that's a separate optimization track and shouldn't all get mixed into this PR.

LiteDB.SourceGenerator/BsonSourceGenerator.cs
```diff
@@ -0,0 +1,1146 @@
...
+    {
+        Scalar,
+        StringList,
+        StringArray,
+        DynamicDictionary,
+        DateTimeOffset,
+        NullableDateTimeOffset,
+        Unsupported
+    }
+}
```
JKamsker, line 1146: This file is WAY too large. Split it up in files and namespaces

LiteDB.SourceGenerator/BsonSourceGenerator.cs
```diff
@@ -0,0 +1,1146 @@
...
+    private static string GetTypeName(ITypeSymbol symbol)
+    {
+        return symbol.WithNullableAnnotation(NullableAnnotation.None)
+            .ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat.WithMiscellaneousOptions(SymbolDisplayMiscellaneousOptions.EscapeKeywordIdentifiers));
+    }
+
+    private sealed record ModelResult(
+        ModelDescriptor? Model,
+        string TypeName,
+        DiagnosticLocationDescriptor? DiagnosticLocation,
```
JKamsker, line 1064: Avoid DiagnosticLocations in a sourcegenerator. Locations potentially change before the sourcegen re-runs. Instead - prefer analyzers

LiteDB.SourceGenerator/BsonSourceGenerator.cs
```diff
@@ -0,0 +1,1146 @@
...
+        string FilePath,
+        TextSpan SourceSpan,
+        LinePositionSpan LineSpan)
+    {
+        public Location Create() => Location.Create(FilePath, SourceSpan, LineSpan);
+    }
+
+    private sealed record ModelDescriptor(
+        string TypeName,
+        ImmutableArray<PropertyDescriptor> Properties,
```
JKamsker, line 1095: Common trap everyone and AI falls into: ImmutableArrays are not incremental!
Fix this and add incrementallity tests (see example: https://github.com/JKamsker/FourSer.Gen/blob/main/tests/FourSer.Tests/IncrementalGeneratorTests.cs)

JKamsker (changes requested):

LiteDB.SourceGenerator/BsonSourceGenerator.cs
zbalkan: Yep, that's why they are immutable. But the question here is, would the properties of a type change in LiteDB? If it happens, shouldn't it be a new record of ModelDescriptor with new properties? It basically means the model has changed, and it will not happen during runtime but through another compilation of the said model.

JKamsker: Thats not the point. The source generator news up a new immutable array with each keypress from the user. So this not being equatable = source gen is not incremental = poor sourcegen performance.

zbalkan: I agree there is an equality issue here, but I would frame it a little differently.

`ImmutableArray<T>` is not inherently incompatible with incremental generators. The actual problem is that `ModelDescriptor` relies on synthesized record equality, while `ImmutableArray<PropertyDescriptor>` does not provide sequence-based equality. If model analysis reruns and produces a new array containing the same descriptors, the resulting `ModelDescriptor` can compare unequal and unnecessarily invalidate downstream steps.

That can reduce the effectiveness of incremental caching, but it does not make the generator itself non-incremental, and Roslyn does not necessarily recreate this value on every user keypress because upstream syntax and semantic steps may themselves be cached.

So I think the appropriate fix is to give the descriptor structural equality for its property collection and add an incremental-generator regression test that verifies unchanged models remain reusable after unrelated edits.

Thanks for pointing this out. The issue is real; I just think the root cause is the equality contract of the intermediate model rather than `ImmutableArray<T>` itself.

PR #2844, opened 2026-09-12
Fix formatted ENSURE error messages

### LINKED ISSUE #2840: ENSURE errors show raw format placeholders ("Page type must be data page: {0}"): the formatted message is discarded
When an `ENSURE` check with arguments fails, the exception message still contains the raw format placeholders, for example `Page type must be data page: {0}`, `invalid segment position in index footer: {0}/{1}` or `Detected loop in FindAll({0})`. The formatted message is built and then thrown away, so the page, index or value that failed never shows up in the error. That is exactly the information needed to diagnose a damaged data file. It affects every `ENSURE(condition, format, args)` call (about 30 across the engine, file reader and disk services).

### Reproduction
Verified on dev `47268cb4`. Damage the page type of one data page and read the collection:

```csharp
using LiteDB;

var file = Path.Combine(Path.GetTempPath(), "ensure-format.db");
File.Delete(file);

using (var db = new LiteDatabase(file))
{
    db.GetCollection("a").InsertBulk(Enumerable.Range(1, 20)
        .Select(i => new BsonDocument { ["_id"] = i, ["s"] = new string('x', 2000) }));
}

// byte 4 of a page header is the page type (Data = 4); mark the first data page as an index page (3)
var bytes = File.ReadAllBytes(file);
var page = Enumerable.Range(0, bytes.Length / 8192).First(p => bytes[p * 8192 + 4] == 4);
bytes[page * 8192 + 4] = 3;
File.WriteAllBytes(file, bytes);

using (var db = new LiteDatabase(file))
{
    try { db.GetCollection("a").FindAll().Count(); }
    catch (LiteException e) { Console.WriteLine($"{e.ErrorCode}: {e.Message}"); }
}
```

Output:

```
999: Page type must be data page: {0}
```

Expected: `Page type must be data page: Index`.

### Cause
The `params` overload of `ENSURE` formats the message and then passes the unformatted `format` to the exception ([Constants.cs:160-162](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Utils/Constants.cs#L160-L162)):

```csharp
var message = string.Format(CultureInfo.InvariantCulture, format, args);

throw LiteException.InvalidDatafileState(format);
```

`LiteException.InvalidDatafileState(string)` uses its argument as the message as-is ([LiteException.cs:356-359](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Utils/LiteException.cs#L356-L359)), so `message` is never used. Affected call sites include:
- [DataPage.cs:17](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Engine/Pages/DataPage.cs#L17) (`Page type must be data page: {0}`)
- [BasePage.cs:285-286](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Engine/Pages/BasePage.cs#L285-L286) (`invalid segment position/length in index footer: {0}/{1}`)
- [IndexService.cs:347](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Engine/Services/IndexService.cs#L347) and the other loop guards (`Detected loop in FindAll({0})`, ...)
- [TransactionService.cs:246](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Engine/Services/TransactionService.cs#L246) (`transaction must be active to commit (current state: {0})`)
- the `FileReaderV8` checks used by `Rebuild()` ([FileReaderV8.cs:159-223](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Engine/FileReader/FileReaderV8.cs#L159-L223))

### Proposed fix
Throw the formatted message:

```csharp
throw LiteException.InvalidDatafileState(message);
```

Optionally guard the `string.Format` call with a `try`/`catch (FormatException)` that falls back to `format`, so a bad format string can never replace the original failure.

Test: call `Constants.ENSURE(false, "value {0} / {1}", 7, "x")` and assert that the `LiteException` has `ErrorCode == LiteException.INVALID_DATAFILE_STATE` and the message `value 7 / x`. The repro above also works as a regression test (expect `Page type must be data page: Index`).

### Related
- #2545: `DropCollection` failing with the literal `page type must be data page: {0}`
- #2827, #2582: `Rebuild()` failing with the literal `Detected loop in FindAll({0})`, which hides which index was affected

NspxMiguel:
`Constants.ENSURE(bool, string, params object[])` now passes its formatted message to `InvalidDatafileState`, so diagnostics contain the failing page, index, or value instead of literal placeholders such as `{0}`.

A focused regression test covers multiple arguments and verifies both the error code and final message.

Fixes #2840.

Validation:
- `dotnet test LiteDB.Tests/LiteDB.Tests.csproj -f net8.0 --filter FullyQualifiedName~Constants_Tests` (1 passed)
- `dotnet test LiteDB.Tests/LiteDB.Tests.csproj -f net8.0 --no-restore --no-build` (346 passed, 7 skipped)
- `dotnet format LiteDB.sln whitespace --no-restore --verify-no-changes --include LiteDB/Utils/Constants.cs LiteDB.Tests/Utils/Constants_Tests.cs`

### DIFF
--- a/LiteDB.Tests/Utils/Constants_Tests.cs
+++ b/LiteDB.Tests/Utils/Constants_Tests.cs
@@ -1,0 +1,18 @@
+using FluentAssertions;
+using Xunit;
+
+namespace LiteDB.Tests.Utils
+{
+    public class Constants_Tests
+    {
+        [Fact]
+        public void Ensure_Formats_Message()
+        {
+            var exception = Assert.Throws<LiteException>(() =>
+                Constants.ENSURE(false, "value {0} / {1}", 7, "x"));
+
+            exception.ErrorCode.Should().Be(LiteException.INVALID_DATAFILE_STATE);
+            exception.Message.Should().Be("value 7 / x");
+        }
+    }
+}
--- a/LiteDB/Utils/Constants.cs
+++ b/LiteDB/Utils/Constants.cs
@@ -159,7 +159,7 @@ namespace LiteDB
 
                 var message = string.Format(CultureInfo.InvariantCulture, format, args);
 
-                throw LiteException.InvalidDatafileState(format);
+                throw LiteException.InvalidDatafileState(message);
             }
         }
 


JKamsker (approved):
Looks good: the formatted message is now thrown instead of the raw format string, which fixes the literal `{0}` placeholders in every ENSURE failure. Small, well-tested change. Thanks!

JKamsker:
Merged, thank you @NspxMiguel! This fixes #2840: ENSURE failures now show the actual page, index or value instead of `{0}`, which makes corruption reports a lot easier to diagnose. Nice, focused fix with a regression test to go with it. Much appreciated, and contributions like this are always welcome.

Issue #2804, opened 2026-09-12
LiteDB crashes on iOS and other AOT-only runtimes (Mono full-AOT, Unity IL2CPP) because it compiles expression trees at runtime


JKamsker:
On AOT-only runtimes, LiteDB crashes on first use. Affected are .NET for iOS / MAUI on iOS, Xamarin.iOS, Unity IL2CPP (iOS, WebGL) and Godot on iOS. The first query, or the first mapping of a type, compiles LINQ expression trees into delegates of LiteDB's own delegate types at runtime, and those runtimes cannot create such delegates without Reflection.Emit. The crash has been reported on Unity since v4, and for 5.0.11 on .NET 6 iOS through .NET 8, .NET 9 and MAUI 9 (January 2026). Dev (`47268cb4`) is unchanged in this respect.

### Symptoms
- **.NET for iOS / MAUI**, from the `BsonExpression` type initializer:
  ```
  Attempting to JIT compile method '(wrapper dynamic-method) LiteDB.BsonValue object:Thunk1ret_BsonValue_IEnumerable`1_BsonDocument_BsonValue_Collation_BsonDocument (...)' while running in aot-only mode.
   (System.ExecutionEngineException)
     at System.Reflection.Emit.DynamicMethod.CreateDelegate(Type , Object )
     at System.Dynamic.Utils.DelegateHelpers.CreateObjectArrayDelegateRefEmit(Type , Func`2 )
     at System.Linq.Expressions.Interpreter.LightLambda.MakeDelegate(Type )
     at System.Linq.Expressions.Expression`1[[LiteDB.BsonExpressionScalarDelegate ...]].Compile()
     at LiteDB.BsonExpression.Compile(BsonExpression expr, ExpressionContext context)
     ...
     at LiteDB.BsonExpression..cctor()
  ```
- **Unity IL2CPP on iOS:** `NotSupportedException: ... System.Reflection.Emit is not supported`, from `Expression.Compile()` in `Reflection.CreateGenericGetter` ← `BsonMapper.BuildEntityMapper` (4.1).
- **Unity WebGL:** `RuntimeError: memory access out of bounds` in `LightLambda.MakeDelegate` under `Reflection.CreateGenericGetter`.
- **Godot 4.4 on iOS:** LiteDB fails to instantiate types while deserializing `List`/`Dictionary` properties, even with `RegisterType`. The same build works on desktop and Android.
[...]

Issue #2832, opened 2026-09-12
Feature roadmap: Source generation, AOT/trimming, package size and serialization speed (generated mappers, AOT annotations, expression interpreter)


JKamsker:
A recurring set of requests is about what LiteDB costs at runtime and at deploy time, not about new features: startup and serialization speed, compatibility with ahead-of-time compilation and trimming, and the size of the package. All of them come back to the same design point. `BsonMapper` discovers types with reflection and builds getters, setters and constructors by compiling expression trees at runtime, and the query layer compiles `BsonExpression`s the same way. That is flexible, but it costs time on first use, it breaks on AOT-only runtimes, and it makes the library hard to trim. The common ask is a compile-time path, generated mapping code, alongside the reflection-based one. This issue collects those requests.

### Requests
- #1667 — opt-in Roslyn source generators for mapping and expression trees, keeping reflection as the fallback
- #2334 — use MessagePack or MemoryPack to speed up serialization and reduce RAM and disk use
- #1826 — split the SQL layer into a separate assembly (`LiteDB.dll` + `LiteDB.Sql.dll`) to reduce size

### Current state on dev
- **Reflection and runtime compilation:** the mapper builds constructors, getters and setters with `Expression.Lambda(...).Compile()` ([Reflection.Expression.cs#L20](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Client/Mapper/Reflection/Reflection.Expression.cs#L20), [#L42](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Client/Mapper/Reflection/Reflection.Expression.cs#L42), [#L80](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Client/Mapper/Reflection/Reflection.Expression.cs#L80)). LINQ translation evaluates captured values the same way ([LinqExpressionVisitor.cs#L723](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Client/Mapper/Linq/LinqExpressionVisitor.cs#L723)), and every `BsonExpression` is compiled ([BsonExpression.cs#L365](https://github.com/litedb-org/LiteDB/blob/47268cb4/LiteDB/Document/Expression/BsonExpression.cs#L365)). The repository has no source generator project.
[...]

Issue #1644, opened 2020-04-20
[BUG] Query.Or is very slow


pejtersen:
**Version**
5.0.7 (Windows 10, .NET Core 2.2)

**Describe the bug**
Performance degradation between 4.1.4 and 5.0.7. `Query.Or` is now much slower.

**Code to Reproduce**
```csharp
var stopwatch = Stopwatch.StartNew();

var query = Query.EQ("field", "value");
for (var i = 0; i < 100; i++)
{
    query = Query.Or(query, Query.EQ("field", i.ToString()));
}

Console.WriteLine(stopwatch.ElapsedMilliseconds); // Takes 2ms in 4.1.4 and 2503ms in 5.0.7
```

holyping:
Or and And operation, it will compiles whole expression to string, concat them with and/or and recompiles them from string to expression object, that's the reason it runs so slow.

PR #2907, opened 2026-09-17
Make LiteDB Native AOT ready

JKamsker:
Native AOT and trimming cannot safely discover arbitrary application models or generate code at runtime. This PR adds explicit warnings on reflection-based APIs, an interpreted expression path that works without dynamic code, and opt-in source-generated typed collections. The document API, SQL, engine features, and string-ID file storage remain usable without runtime model mapping.

The source generator is a separate package, paired with the LiteDB runtime version. Its supported model subset and compatibility boundaries are documented in [docs/AOT.md](https://github.com/JKamsker/LiteDB/blob/codex/pr-2766-aot-validation/docs/AOT.md) and [the package contract](https://github.com/JKamsker/LiteDB/blob/codex/pr-2766-aot-validation/docs/SourceGeneratorPackage.md). Builds are based on current `dev` (`7d2a16c43`); the SQL `REBUILD` fix is already upstream in #2915.

## Merge-confidence assessment — 2026-09-18

**Strong evidence for the documented AOT paths; a conditional merge recommendation, not a claim that every LiteDB API is AOT-safe.** The concrete mapping, warning-boundary, packaging, and test-coverage defects found in this pass are fixed in `f528a7551`, `4e574224d`, and `7662d3a0e`. **All 57 CI jobs pass on the final head**, including all ten Native AOT jobs. Keep the PR draft until the compatibility/API decisions, publication authority, and maintainer review below are resolved. Green automation does not replace that review: there are no submitted approving reviews, and CodeRabbit's successful status currently means it skipped this draft.

### Questions answered with code or tests

| Question | Answer and evidence |
| --- | --- |
| Is the branch current, and does ordinary LiteDB still work? | `dev` is an ancestor of the PR; local and remote heads match. Local .NET 8 suite: **1,021 passed, 7 skipped**. Final CI: **57/57 successful jobs**, including ordinary tests on Linux, macOS, Windows, and .NET Framework 4.6.2/4.8.1. See the ordinary matrix-label qualifications below. |
| Do supported paths actually run after trimming and Native AOT publishing? | Four-mode gate: regular, trimmed, Native AOT, and whole-library-rooted Native AOT. Every `ILxxxx` publish diagnostic fails the gate; observed-value transcripts must match. **All ten final-head CI AOT jobs pass** across Linux x64/arm64/musl, Windows x64, and macOS arm64 on .NET 8 and 10, with no IL diagnostics in the supported published scenarios. |
| Are important engine features covered? | Published transactions/reopen, encryption, shared mode, parallel writers, rebuild/pragmas, culture collation, vector search, document/generated LINQ, multi-chunk file storage; expression-method/operator completeness sweep and SQL statement/clause/system-collection sweep. Existing CI also checks v8/v9 vector-file compatibility with LiteDB 5.0.21. |
| Can generated operations secretly fall back to runtime model discovery? | Guarded smoke mappers throw from the runtime mapping hooks. Generated query paths retain their registered deserializer. Managed diagnostics, operation, serialization, and incrementality coverage: **91 tests passed**. |
| Do inherited IDs have the same stored shape as reflection mapping? | **Fixed.** `Person.PersonId` inherited by `Employee` previously produced an unrelated generated `_id`. Regression tests now cross-read both paths, compare stored BSON, check LINQ and auto-ID assignment, and test `[BsonId]` → `Id` → `<DeclaringTypeName>Id` precedence. Published smoke coverage exercises inherited IDs too. |
| Can unsupported attributes or duplicate BSON fields silently lose data? | **Fixed two gaps.** `[BsonRef]` on an otherwise supported property now reports `LDBSG002`; field names differing only by case now report `LDBSG003`. Inherited getter-only conventional IDs are rejected rather than silently omitted. |
| Can several generated model libraries coexist? | **Fixed.** Registration classes are internal to each assembly; libraries expose their own public registration wrapper. A compiler regression test composes two model libraries and an application with type-conflict warnings promoted to errors. |
| Must existing `ILiteDatabase` implementations add a member? | **No longer.** `GetGeneratedCollection<T>` stays on `LiteDatabase`. A fixture implementing the pre-PR interface compiles and performs upload/read/download/delete through `LiteStorage`'s original typed-collection fallback. |
| Are arbitrary custom file IDs safe without warnings? | **No, and the API now says so.** A real trimmed probe collapsed two nested IDs into one file (`Home:{}`), with no diagnostics before this fix. `LiteDatabase.GetStorage<T>`, `ILiteDatabase.GetStorage<T>`, and the public `LiteStorage<T>` constructor now emit both `IL2026` and `IL3050`; a package-level negative compilation test requires all six diagnostics. Default `FileStorage` remains warning-free. Flat custom IDs are separately qualified by narrowly justified smoke-test suppressions. Ignoring the warnings does not make nested IDs safe. |
| Do existing native-type file-ID converters still work? | **Fixed.** Serialization now respects registered converters before BSON-native conversion. A previously failing string-ID converter test and a published smoke scenario verify storage and round-trip lookup. |
| Does persisted type-name lookup surface its trimming risk? | **Fixed.** The public `DefaultTypeNameBinder.GetType` suppression hid that a stored type name may refer to a removed type. Both it and `ITypeNameBinder.GetType` now carry `RequiresUnreferencedCode`; package-level negative compilation checks require warnings for direct and interface calls. |
| Can correctly named packages contain validation DLLs with the wrong version? | **Prevented.** A dry run reproduced `6.0.0`-labelled packages containing `0.0.0.0` DLLs. Synthetic consumer builds now use `PackageValidation`; release pack rebuilds production binaries. Archive and embedded-assembly-version checks run before publication, and Linux CI dry-runs them without publishing. Workflows now read GitVersion 6's `SemVer` field and disable `UseFullSemVerForNuGet` so nuspec and filename versions agree. |
| Can a crashed retry make the AOT suite gate falsely green? | **Fixed.** Retry results are cleared, exit status is checked, and only an explicit `Pass` qualifies. Seven black-box gate tests cover pass, fail, crash, missing result, skip, a pass record followed by a crash, and visibility of shared baseline failures. |
| Does encrypted legacy-file rejection/upgrade actually execute in the published test host? | **Now yes.** A fixture lookup previously failed in both JIT and Native AOT and therefore escaped the difference gate. The corrected resource lookup passes in both modes and in xunit. The gate now prints every JIT baseline failure by name and reason so shared failures remain visible. |
| Does the entire reflection-based test suite pass as Native AOT? | **No.** The separate host compares individual outcomes with JIT and permits only named, categorized AOT-only failures. Final local Linux/.NET 8 results: JIT **1,021 passed, zero failed**; Native AOT **957 passed, 64 known failures**; seven skipped in each. Reflection-generated generic types and test/assertion-library reflection remain outside the supported contract. Failures shared with JIT are reported separately, not counted as AOT regressions; ordinary xunit CI is a separate gate. Passing the difference gate is not equivalent to every test passing natively. |

### Decisions and evidence still needed

- [ ] **Approve the remaining binary/source compatibility changes for the intended release.** `BsonMapper.ResolveCollectionName` is a property instead of a public field: precompiled consumers need rebuilding. Overrides/interface implementations may need matching trim/AOT attributes. Generated collection support no longer adds an interface member, but the field-to-property change still needs explicit major-version/release-note acceptance.
- [ ] **Approve the generator's release scope before freezing its public API.** It handles sealed, top-level mutable models and a documented subset of collections. Nested entities, arbitrary `List<T>`, constructor/init records, DbRefs, and .NET Framework/netstandard2.0 generated consumers are not supported. Unannotated getter-only projections are omitted. Mapper customization remains conservatively rejected globally, including unrelated `RegisterType`/fluent mapping changes; per-model compatibility is a follow-up. Decide whether this is acceptable for the release, should be labelled preview, or should be split from the annotation work. Recommendation: explicitly label the generator API as preview if retained; this recommendation is not an accepted release decision.
- [ ] **Confirm NuGet publication authority for both package IDs before merging.** Packing/restoring/running both packages is tested. API-key scope for `LiteDB.SourceGenerator` cannot be proved by local packing; no credentials were retrieved and no package was published. Pushes to `dev` trigger prerelease publishing, so this matters at merge time, not only at a later stable release. A release owner must confirm scope; publishing the pair is not atomic.
- [ ] **Complete supported-toolchain mobile qualification if it is a merge requirement.** Prior recorded runs passed .NET iOS Mono full AOT (`UseInterpreter=false`), iOS NativeAOT, and Unity 6 IL2CPP on simulator/device. They used Xcode 27 with a workload expecting 26.6 and are historical evidence, not a rerun of this head. Still open: supported Xcode/workload pairing, exact before/after of #2804's app, App Store/TestFlight signing, other Apple targets/older OS versions, and Unity generated models/minimal preservation/stress coverage. See [the reproducible mobile report](https://github.com/JKamsker/LiteDB/blob/codex/pr-2766-aot-validation/docs/ios-aot-validation.md). Unity's default runtime probe also does not select the new path automatically; public opt-in is undecided.
- [ ] **Obtain maintainer review of the final contract and diff.** The earlier [review](https://github.com/litedb-org/LiteDB/pull/2907#issuecomment-5712412956) was valuable, but is not an approval of the current changes. Final CI is green; there are still no submitted approving reviews.

### Compatibility and evidence qualifications

- Queries on `BsonDocument` collections now reject captured application objects that would require reflection. This also affects ordinary JIT applications. Convert values to BSON first; ordinary typed collections retain runtime mapping.
- Generic custom-ID storage warns conservatively even for scalar IDs or registered converters. Use default `FileStorage`, or qualify the complete ID conversion in a published application before narrowly suppressing warnings. The top-level preservation annotation does not preserve nested models. A separate Linux/.NET 8 probe verified the documented complete nested-ID converter in both trimmed and Native AOT modes: two distinct IDs and their original payloads survive, with no IL diagnostics under narrowly justified method suppressions.
- Object initializers/anonymous types in expression trees can produce consumer-side `IL2026`; invariant globalization cannot open culture-collated files. Published tests do not remove those platform/runtime constraints.
- The pre-existing ordinary CI matrix overstates some distinctions: Windows `x86` rows interpolate the architecture into labels/log names but do not select an x86 SDK or test host, and `.NET 9` rows still target `net8.0` with .NET 8 installed. Treat those as duplicate/SDK-oriented checks, not proof of actual x86 or .NET 9 runtime coverage. Correcting that matrix and asserting the executing process architecture/runtime is follow-up validation work. The Native AOT gate independently publishes explicit RIDs on .NET 8 and 10; its platform claims do not rely on those ordinary rows.
- Existing Windows/.NET 8 measurements show roughly 1.2–2× slower warm expression scans under AOT and much faster first-use expression setup. These are workload-specific historical measurements, not performance guarantees for this head.
- The synthetic consumer version `0.0.0-sourcegenerator-p13.1` triggers NuGet's existing `<5.0.13` advisory match ([GHSA-3x49-g6rc-c284](https://github.com/advisories/GHSA-3x49-g6rc-c284)). This is visible and unsuppressed; it is not a new analyzer-package vulnerability or a substitute for release security review.

## Review cleanup

The applicable “What to cut” items are addressed: unshipped DateTimeOffset compatibility, dead generated delegates/API members, runtime-project generator reference, duplicate compile/snapshot/smoke scaffolding, tiny model files, one-off measurements, and development jargon. The package consumer is a small round trip plus a targeted API-warning probe. The LINQ file split remains because recombining it would exceed the repository's existing non-growing C# size limit. Stale documentation about the removed compatibility reader, snapshots, and generator targets was corrected.

## Validation

- `dotnet test LiteDB.Tests/LiteDB.Tests.csproj -f net8.0 -c Release -p:TestingEnabled=true --settings tests.runsettings` — 1,021 passed, 7 skipped.
- `dotnet test LiteDB.AotTests/LiteDB.AotTests.csproj -c Release --settings tests.runsettings` — 91 passed.
- `scripts/validate-source-generator-package-consumer.sh` — package archive, required negative diagnostics, trimmed execution, and Native AOT execution passed.
- `TARGET_FRAMEWORK=net8.0 scripts/validate-aot-feature-parity.sh` — all four modes passed, identical transcripts, zero IL diagnostics.
- `TARGET_FRAMEWORK=net8.0 scripts/validate-aot-test-suite.sh` — JIT 1,021 passed / zero failed; Native AOT 957 passed / 64 explicitly allowlisted failures; seven skipped in each. Encrypted legacy-file rejection/upgrade passes in both modes.
- `python3 scripts/test-aot-test-suite-gate.py` — all seven failure-handling/reporting cases passed.
- Release-pair dry run — passed: both nuspecs match their filenames; all three runtime target DLLs and the analyzer have the expected `AssemblyVersion=6.0.0.0`. The new guard rejects the deliberately stale `0.0.0.0` packages.
- [Full CI for `7662d3a0e`](https://github.com/litedb-org/LiteDB/actions/runs/35320548159) — **57/57 jobs successful**, including all ten Native AOT jobs, package-consumer and production-package validation, legacy .NET Framework tests, cross-process tests, and plain/encrypted vector-file compatibility with LiteDB 5.0.21. Every custom-host JIT baseline passes 1,021 tests; Native AOT failures match the documented allowlist. CodeRabbit's separate successful status is a skipped draft review, not an approval.

No merge, release publication, or external review request has been performed.


JKamsker:
# Review: AOT annotations + source generator (#2766 plus the validation commits here)

**Summary:** I don't think this should merge as one PR. The annotation half is ready and clearly worth having. The source-generator half works, but it covers too small a set of models to commit to as public API yet. Roughly a quarter of the diff can go. AOT compatibility is proven for a narrow set of paths and only asserted (through suppressions) for the rest.

| Question | Answer |
|---|---|
| Complete? | For what it claims, mostly yes: CI is green, including the Native AOT job. As a feature, no. |
| Too much? | Yes. +9,859 lines in 174 files, of which about 6,100 lines are tests covering the same scenarios four times. |
| Can it be leaner? | Yes. Roughly 2,500 lines and 60 files can go with no loss of coverage. |
| AOT fully proven? | No. Partly proven, partly asserted through suppressions. |
| Good addition? | The direction is good. The annotations are worth merging now. The generator as shaped is not yet. |

This review was done with Claude Code as an AOT-focused second reviewer, as @zbalkan suggested on #2766. Everything under "What was run" was actually executed; items marked *unconfirmed* come from reading the code only.

## What was run

- **Windows Native AOT:** the smoke suite publishes and passes on win-x64 (7.7 MB executable). CI only covers linux-x64.
- **LINQ under AOT:** a scratch AOT consumer using closures, `StartsWith`, `Select(x => x.Name)`, `GroupBy` and `EnsureIndex(x => x.Age)` published with zero IL warnings and ran correctly. The smoke suite does not exercise these.
- **Package dependency:** packing `LiteDB` yields a nuspec with no dependency on `LiteDB.SourceGenerator` and no analyzer entries, so the analyzer `ProjectReference` in `LiteDB.csproj` does not leak. It is still unnecessary (see below).

## Is AOT proven?

What is actually proven: generated CRUD on flat models, the `BsonDocument` API, simple SQL, and basic LINQ, on linux-x64 and win-x64, net8, with invariant globalization.

The gaps:

1. **Blanket suppressions on `BsonMapper` (`BsonMapper.cs:23-27`).**
   - Five IL codes are suppressed at class level across every partial.
   - The justification strings are the `RequiresUnreferencedCode` messages reused, so they describe why the code is *unsafe* rather than why the suppression is safe.
   - Public `Entity<T>()`, `GetExpression` and `GetIndexExpression` reach reflection with no annotation.
   - Any future regression inside the class is silenced as well.
2. **"Never falls back to runtime member discovery" (`docs/AOT.md:132`) is not accurate.**
   - `LinqExpressionVisitor.cs:281` calls `_mapper.Serialize` on captured values.
   - `LinqExpressionVisitor.Member.cs:27-29` falls back to `GetEntityMapper` for types without a generated map.
   - `FailOnGenericConversionMapper` only guards `ToDocument` and `ToObject`, so it would not catch either fallback.
3. **The warning gate is weaker than the docs suggest.**
   - The scratch consumer used exactly the documented setup (`ILLinkTreatWarningsAsErrors`) and published successfully *with* IL2026 warnings, so that property does not gate the ILC step.
   - Only the nine IL codes listed explicitly in the smoke csproj's `WarningsAsErrors` become errors.
4. **Two consumer-facing gaps are undocumented.**
   - `UpdateMany(x => new T {…})` and `Select(x => new T {…})` raise IL2026 in the *consumer's* code (from `Expression.Bind`). That is why the smoke test suppresses IL2026 around its own queries. It works at runtime.
   - One unrelated `RegisterType<Version>(…)` or `Entity<X>()` call on the mapper makes **every** generated collection throw (`BsonMapper.GeneratedExecution.cs:98-99`). `RegisterType` is otherwise the AOT-safe way to extend the mapper, so many real apps would hit this immediately.
5. **Never run under AOT:** file storage, encryption, shared mode and mutexes, `Rebuild`, transactions, vector search, real collation (everything runs with `InvariantGlobalization=true`), and net10.
6. **The transcript diff proves little.** Every printed line is a fixed string, so comparing the three transcripts byte for byte only shows that none of the three publish modes threw.

## Bugs and design problems

- **Silent ID divergence (high).**
  - The reflection mapper picks the conventional ID as `DeclaringType.Name + "Id"` (`BsonMapper.GetEntityMapper.cs:183`).
  - The generator uses the annotated type's name (`BsonModelAnalyzer.cs:146`, and again in `IsComputedProperty`).
  - A base-class `PersonId` on a sealed `Employee` is therefore `_id` on one path and an ordinary field on the other.
  - The generator reports no diagnostic, so the two paths produce mutually unreadable data.
- **`ILiteDatabase.GetGeneratedCollection<T>` breaks implementers.** Adding a member to the interface breaks every external implementer, decorator and mock, and netstandard2.0 has no default interface methods to soften that. Put it on `LiteDatabase` only, or make it an extension method.
- **`public static class LiteDB.Generated.LiteDbGeneratedMappings` has a fixed name in every consuming assembly.** Two libraries that both use the generator cannot be referenced together (CS0433). Make it `internal` or derive the name from the assembly.
- **The supported subset is too narrow.** No nested objects, no `List<T>` beyond strings, no init or constructor records, sealed types only. `PropertyKind.StringList` / `StringArray` are hardcoded, so nested types need a redesign of the Models and Emission layers rather than an extension.
- **Getter-only properties are silently dropped,** where the reflection mapper persists them. Documented, but there is no diagnostic.
- **Generated code does not compile for netstandard2.0 / net4x consumers** (e.g. `Enum.Parse<T>`), and there is no diagnostic for that.

On the positive side: generator incrementality is correct and tested with tracked steps, the models hold no Roslyn objects, the emitted code is clean (`global::` names, escaped identifiers and literals), and the on-disk format matches the reflection mapper for everything the subset supports.

## What to cut

- **Compat code for formats that never shipped.** The `{DateTime, Offset}` DateTimeOffset reader (also added to the ordinary `BsonMapper`) and the docs' reference to "the former `LiteAotDatabase`" both exist only in this PR's own history, not on `dev`.
- **Dead emitted delegates** (`SerializeStringList`, `CreateInstance`, per-member `Serialize`/`Deserialize`/`Getter`): the generated-path runtime only reads metadata and the id `Setter`.
- **Dead API surface:** the option-less `GeneratedEntityMap` constructor, `SupportsScalarOptions`, the always-true `CanEmitExecutionMap`, the `IsDbRef` check that cannot fire.
- **The generator reference in `LiteDB.csproj`.** LiteDB itself uses no `[BsonSourceGenerated]`.
- **`LiteDB.SourceGenerator.PackageConsumer`:** shrink from 423 to about 60 lines. Much of it is a copy of the smoke tests, down to identical GUIDs. It only needs to prove the nupkg restores, the analyzer loads and one round trip works.
- **`LiteDB.SourceGenerator.CompileTests`:** redundant with `LiteDB.AotTests` (add CS8619 to its `WarningsAsErrors` instead).
- **`LiteDB.AotTests/Models`:** 45 files with a median of 10 lines. Consolidate to about 6.
- **`GeneratedSourceSnapshotTests.cs` (384 lines):** pins emitter-local names like `map0`, so any rename breaks it. Replace with a few real snapshots or cut.
- **Boundary scenarios in the smoke suite:** they belong in the managed tests only; the smoke suite needs one scenario per feature family.
- **`measure-source-generator-incremental-build.sh` and its doc (310 lines):** one-off measurements from one machine.
- **The output-root safety logic in `validate-aot-feature-parity.sh`.**
- **Drive-by refactors:** `_mapper` → auto-property, `[]` collection expressions, csproj whitespace, the `LinqExpressionVisitor` file split, and `ToLower() ==` → `OrdinalIgnoreCase` (a quiet behaviour change).
- **Development jargon:** C1/C2/PhaseB/PhaseC appear about 450 times, the hint name is `…v2.g.cs`, and the docs say things like "first mapping slice".

Separately, `LiteDB.AotTests` uses MSTest while the rest of the repo uses xunit.

## Release-pipeline risks (*unconfirmed*)

- Both publish workflows now push a second package to nuget.org, so the API key must be scoped for the new `LiteDB.SourceGenerator` ID. Otherwise `LiteDB` gets pushed and the step then fails.
- The consumer script runs `dotnet pack … -p:GitVersionEnabled=false` before the real `dotnet pack --no-build`. That may rebuild the shipped DLLs with `AssemblyVersion=0.0.0.0`. This needs a real dry run before anyone trusts it.

## Recommendation

Split into three PRs:

- **PR A, merge soon (about 600 lines).**
  - The `Requires*` annotations, the AOT/trim analyzers, IL warnings as errors, `Compile(preferInterpretation: true)`, and a smoke test covering only the `BsonDocument` API.
  - This gives every AOT/trimming user correct warnings today, at low risk.
  - First replace the class-level `BsonMapper` suppression with member-level ones that carry real justifications, and annotate `Entity<T>` and `GetExpression`.
  - Mention `ResolveCollectionName` changing from a public field to a property (a binary break) in the release notes.
- **PR B, the generator.**
  - Ship it as experimental/preview, with a minimal public surface.
  - Fix the ID convention, the interface break and the `LiteDbGeneratedMappings` name collision.
  - Scope the mapper guard per type instead of per mapper.
  - Before freezing any API, decide whether to move to a recursive type-shape model or build on mrdevrobot's fork (linked from #2832), which already handles nested types and collections.
- **PR C, packaging and release workflows,** reviewed in isolation with a dry run.

@zbalkan's point on #2766 that AOT needs a different kind of review holds up: these problems only surfaced by publishing a real consumer and by reading the suppressions one by one.

JKamsker:
### Proposed AOT consolidation from #2917 / #2908

This PR already addresses #2804's no-dynamic-code expression failure, so I recommend making #2907 the home for that work rather than maintaining a competing AOT implementation in the bugfix stack. The useful transfer is primarily **targeted regression coverage**, adapted to this PR's implementation and supported API contract—not a wholesale cherry-pick.

#### Where to find our work

- [78aa2eb6 — Interpret LiteDB expressions when dynamic code is unavailable (#2804)](https://github.com/litedb-org/LiteDB/commit/78aa2eb6c52ab9182c464794d9f1e79dacddfd1f): the AOT implementation, startup opt-in, forced-interpreter/cache tests, numeric/nullable/evaluation-order tests, and `docs/aot-runtime.md`.
  - [Issue2804ForcedInterpreter_Tests.cs](https://github.com/litedb-org/LiteDB/blob/78aa2eb6c52ab9182c464794d9f1e79dacddfd1f/LiteDB.Tests/Issues/Issue2804ForcedInterpreter_Tests.cs)
  - [Issue2804NumericInterpreter_Tests.cs](https://github.com/litedb-org/LiteDB/blob/78aa2eb6c52ab9182c464794d9f1e79dacddfd1f/LiteDB.Tests/Issues/Issue2804NumericInterpreter_Tests.cs)
  - [RuntimeExpression.cs](https://github.com/litedb-org/LiteDB/blob/78aa2eb6c52ab9182c464794d9f1e79dacddfd1f/LiteDB/Utils/RuntimeExpression.cs) and [runtime notes](https://github.com/litedb-org/LiteDB/blob/78aa2eb6c52ab9182c464794d9f1e79dacddfd1f/docs/aot-runtime.md).
- [0ac1ccb5 — Defer boolean compilation until execution (#1644)](https://github.com/litedb-org/LiteDB/commit/0ac1ccb52e544701582b8bf76d963b162570fbf3): a later dependency that extends the AOT helper and tests preservation of the selected execution mode across deferred compilation. The general deferred-compilation fix belongs in #2917; this is a reference for integration work, not a recommendation to move that whole commit here.

#### Coverage worth carrying over

1. **Short-circuiting and evaluation order:** throwing branches must remain unexecuted under `AND`, `OR`, and `IIF`; captured calls/indexers must be evaluated exactly once, with normal CLR argument-evaluation order.
2. **Captured-value conversion boundaries:** representative nullable/coalescing, enum, checked-overflow, unboxing, and lifted-conversion cases exercised through LiteDB's LINQ translation. The full numeric matrix primarily validates our custom interpreter; adapt useful cases rather than importing a test suite for an implementation this PR does not use.
3. **Proof that AOT-path tests cannot silently reuse JIT delegates:** our forced-mode test checks cache isolation. This PR already uses unique expression strings; strengthen that with a cold-process scenario and add the important semantic cases to the existing published-AOT validation. Managed forced-mode tests alone do not prove NativeAOT or mobile compatibility.
4. **Selected end-to-end combinations:** captured values with indexed queries, nested `MAP` projections, and nullable/array round trips, only where they add coverage beyond the existing smoke suite. Keep tests within this PR's documented mapping contract; do not imply arbitrary reflection-mapped models are trimming-safe.

#### Decisions before carrying behavior over

- **Startup opt-in:** consider the capability behind `AppContext.SetSwitch("LiteDB.UseInterpreter", true)` for Unity/older runtimes whose dynamic-code detection is insufficient. If accepted, adapt it to this PR's compiler and verify in a fresh process that it takes effect before any compilation. This is an API/platform decision, not a reason to import our interpreter.
- **Captured-evaluation exceptions:** our tests retain `TargetInvocationException` wrapping, whereas this PR invokes the captured delegate directly. Decide the intended compatibility contract before copying the assertion.
- **Deferred compilation:** retain #1644 in the bugfix PR, decouple it from the removed helper, and verify the combination with this PR when integrating the branches.

I would **not** transfer `RuntimeExpression`/`ExpressionInterpreter` wholesale, impose our custom interpreter's restriction on inline captured selector lambdas, or classify application reflection-metadata preservation as a new product bug. These are code-review transfer candidates, **not newly demonstrated failures in #2907**. No code has been moved by this comment.

JKamsker:
@zbalkan can i get your opinion on this? I prefer this one here since it is a little cleaner

Issue #2926, opened 2026-09-18
Flaky test: Index_With_Like depends on the culture Collation.Default was initialised under


JKamsker:
**Test:** `LiteDB.Tests.Engine.Index_Tests.Index_With_Like`

**Symptom:** fails intermittently with

```
Expected r3.Length to be 4, but found 3.
```

`LIKE 'ma%ci%'` returns 3 of the 4 `mauricio` case variants. A re-run of the same job on the same commit passes.

**Seen on**
- 2026-09-17, PR #2914, `Test (Linux ARM64 - .NET 8)`
- 2026-09-18, `dev` @ `f49b9d75` (right after #2916 merged), workflow *Publish prerelease*, job `Test (Windows windows-2022 - x86 - .NET 9)` — [run 35375488244](https://github.com/litedb-org/LiteDB/actions/runs/35375488244). The `CI` workflow on the same commit was green. Because the test matrix gates packing, **no prerelease package was published for that commit.**

So it is not tied to one OS, architecture or runtime.

**Cause:** the test opens `filename=:memory:` without a collation, i.e. it uses the ambient `Collation.Default`, but asserts English case folding (`I`/`i` equivalent). `Collation.Default` is initialised once per process from the current culture. When a test that switches `CultureInfo.CurrentCulture` to `tr-TR` happens to run before that initialisation, the default collation is Turkish, `I` and `i` no longer fold together, and one variant drops out. Test order varies between runs, hence the flake. A fresh-process probe reproduces the exact 4-vs-3 mismatch when `Collation.Default` initialises under Turkish culture (see #2917). #2916 added more Turkish-culture tests (`Issue2858_Tests`), which plausibly raises the failure rate on `dev`.

No production behaviour is wrong here; the fixture's assumption is.

**Fix:** already written in #2917 (its #2144 commit): the test becomes a theory over explicit `en-US/IgnoreCase` (4 / 4 matches) and `tr-TR/IgnoreCase` (3 / 0 matches) collations, so it no longer depends on the process-wide default.

**Until #2917 merges:** the test is skipped on `dev` so it cannot block prerelease publishing. #2917 re-enables it (its version of the test replaces the skipped one; expect a trivial conflict on the `[Fact]` line when rebasing).

- [ ] skip on `dev`
- [ ] re-enable via #2917

JKamsker:
Skip on `dev`: #2927. Re-enable: #2917 (its version of the test replaces the skipped one).

PR #2927, opened 2026-09-18
test: skip flaky Index_With_Like until #2917 re-enables it

JKamsker:
Skips `Index_Tests.Index_With_Like` on `dev` until its real fix lands with #2917. Tracked in #2926 (deliberately **not** `Closes` — the issue stays open until the test is re-enabled).

## Why

The test fails intermittently with `Expected r3.Length to be 4, but found 3`. It made the *Publish prerelease* run on `dev` red right after #2916 merged ([run 35375488244](https://github.com/litedb-org/LiteDB/actions/runs/35375488244), job `Test (Windows windows-2022 - x86 - .NET 9)`), while `CI` on the same commit was green. Because the test matrix gates packing, no prerelease package was published for that commit.

Cause: the fixture asserts English case folding but uses the ambient `Collation.Default`, which is initialised once per process from the current culture. If a Turkish-culture test runs first, `I`/`i` stop folding together and one of the four `mauricio` variants drops out. Test order varies per run. Production behaviour is not wrong; the fixture's assumption is. Details in #2926.

## What

One line: `[Fact]` → `[Fact(Skip = "…#2926…#2917")]`. No production code.

## Follow-up

#2917 already contains the fix (the test becomes a theory over explicit `en-US/IgnoreCase` and `tr-TR/IgnoreCase` collations). Rebasing #2917 onto this will conflict on exactly this attribute line; #2917's version wins and re-enables the test.

## Test plan

- [x] `Index_Tests` locally on net10.0: 4 passed, 1 skipped (`Index_With_Like [SKIP]`), 0 failed
- [ ] CI green

## Summary by CodeRabbit

- **Tests**
  - Skipped a flaky index matching test that depends on the default English case-folding behavior.
  - The test now references the related issue for tracking.


JKamsker:
Fallback only. #2917 contains the real fix for this test (explicit `en-US` / `tr-TR` collations) and now carries `Closes #2926`. If #2917 merges first, close this PR unmerged. If this one is merged first, #2917 gets a one-line conflict on the `[Fact]` attribute; #2917's version wins and re-enables the test.

JKamsker:
Superseded: #2917 is merged and contains the real fix for `Index_With_Like` (explicit `en-US` / `tr-TR` collations); #2926 is closed by it.

